PR #227, opened 2014-04-13
Created new function "pulseDescriptionFromStatusByte"

amullins83:
Created new function `pulseDescriptionFromStatusByte` to make parsing the LED pulse speed and style descriptions easier to understand. I am not attached to the function name, but it seems like the bitwise operations should be pulled out to a helper function, and since the values are used as indexes anyway, it seems more expressive to explicitly index into an array of descriptions. In a production app, this would make it easier to globalize, assuming these text descriptions would ever be presented to the user. I may be over-thinking this.


rios1223:
To closed the bug

PR #351, opened 2015-01-27
[usb-device-info] Remove deprecated chrome.usb.requestAccess

beaufortfrancois:
According to https://developer.chrome.com/apps/usb#method-requestAccess, we should remove it from the usb-device-info sample.

TBR=@reillyeon

### DIFF
--- a/samples/usb/device-info/manifest.json
+++ b/samples/usb/device-info/manifest.json
@@ -1,6 +1,6 @@
 {
   "name": "USB Device Info",
-  "version": "0.3",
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
Minimum chrome version has been updated. 
Thank you for the reminder.

beaufortfrancois:
Application updated: https://chrome.google.com/webstore/detail/usb-device-info/igkmggljimacfdfalpeelenjeicmfnll

Issue #350, opened 2015-01-15
need an example of scanning with bluetoothLowEnergy


monteslu:
The current Bluetooth Low-Energy Device Information Service Demo requires that the client already knows about the device.  Bluetooth 4.0 (BLE) provides the ability to scan nearby devices based on a service UUID, and then connect and query characteristics and descriptors.

XhuaChen:
+1

sandeepmistry:
+1

beaufortfrancois:
The "Start Discovery" button has actually been added previously to this bluetooth sample but not published in the Web Store. 
I've just bumped the manifest app version and updated the webstore item at https://chrome.google.com/webstore/detail/dblimghcclaakknbpajckcamddhjaaai/

Thank you for raising this issue!

monteslu:
The problem is that it is still using bluetooth 3.0 APIs.
Other libraries such as Noble for Node, or iOS and Android BLE libraries allow you to scan by service UUID.

This example app is calling chrome.bluetooth.getDevices() which only lists things that are already connected or saved to the OS.   As far as I can tell, the discovery toggle handler in the code only logs errors if it finds them..

monteslu:
Should reopen this issue

beaufortfrancois:
I'm not sure about BT 3.0 APIs but from what I can read at https://developer.chrome.com/apps/bluetooth#method-startDiscovery,

_Start discovery. Newly discovered devices will be returned via the onDeviceAdded event. Previously discovered devices already known to the adapter must be obtained using getDevices and will only be updated using the |onDeviceChanged| event if information about them changes._

Maybe @oahziur can help us there.

beaufortfrancois:
You're actually right, the battery-service-demo and heart-rate-sensor bluetooth samples use onDeviceAdded to gather new discovered devices.
Therefore, I will add it.

Thank you Luis!

beaufortfrancois:
Hey Luis,
I've just created a Pull Request https://github.com/GoogleChrome/chrome-app-samples/pull/353 that you might want to try before it's getting reviewed.
Please let me know what you think.

beaufortfrancois:
PR has been accepted and Webstore Item published at https://chrome.google.com/webstore/detail/device-information-servic/dblimghcclaakknbpajckcamddhjaaai
Please reopen the issue if there's something wrong.

PR #353, opened 2015-01-30
[device-info-demo] Handle onDeviceAdded and onDeviceRemoved events

beaufortfrancois:
Similar to the battery-service-demo bluetooth sample, the device-info-demo bluetooth sample needs to handle `onDeviceAdded` and `onDeviceRemoved` events to get notified about new discovered devices.

BUG=#350
TBR=@scheib


scheib:
Reading over code this LGTM

Issue #335, opened 2014-11-04
Add chrome app frame colors sample 


beaufortfrancois:
According to https://developer.chrome.com/apps/app_window#type-FrameOptions, we should have an example of `activeColor` and `inactiveColor` in the [Window Options](https://github.com/GoogleChrome/chrome-app-samples/tree/master/samples/window-options) sample

beaufortfrancois:
Closing with this

PR #358, opened 2015-02-09
Add color to window options

beaufortfrancois:
Here's an example of `activeColor` and `inactiveColor` for the Window Options sample

screenshot 2015-02-09 at 10 10 53 am

BUG=#335
TBR=@scheib

### DIFF
--- a/samples/window-options/manifest.json
+++ b/samples/window-options/manifest.json
@@ -1,7 +1,7 @@
 {
   "manifest_version": 2,
   "name": "Window Options Sample",
-  "version": "1",
+  "version": "2",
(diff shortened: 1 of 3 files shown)


scheib:
lgtm, though changes like this should go to folks near benwells who authored the feature.

beaufortfrancois:
Thank you Vincent.
Sorry for that. I will be more careful next time.

beaufortfrancois:
Updated Webstore Item at https://chrome.google.com/webstore/detail/window-options-sample/cfcgoifcnpnadlhhoolkemkjkhoajfmk

PR #361, opened 2015-02-12
Update index.js

jordangraft:
Fixed typo in 'response_type'

### DIFF
--- a/samples/github-auth/index.js
+++ b/samples/github-auth/index.js
@@ -26,7 +26,7 @@ var gh = (function() {
         var options = {
           'interactive': interactive,
           url:'https://github.com/login/oauth/authorize?client_id=' + clientId +
-              '&reponse_type=token' +
+              '&response_type=token' +
               '&access_type=online' +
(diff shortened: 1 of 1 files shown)


beaufortfrancois:
Hello Jordan!

According to https://developer.github.com/v3/oauth/#redirect-users-to-request-github-access, `response_type` is not required anymore so we might simply get rid of it.

I see that `access_type` is not mentioned either...

What do you think about this?
```javascript 
var options = {
          'interactive': interactive,
          url:'https://github.com/login/oauth/authorize?client_id=' + clientId +
              '&redirect_uri=' + encodeURIComponent(redirectUri)
        }
```

jordangraft:
Sounds great to me.  Thanks for taking a look at it and thanks for building this example.  It was extremely invaluable as a template for integrating with our own app's oauth endpoint and our chrome extension.

beaufortfrancois:
Thanks Jordan!
I've pushed #362 and bumped manifest version.

The Webstore Item has also been updated at https://chrome.google.com/webstore/detail/laolmfhjaobpboigjfbclcphckmjodlp

PR #347, opened 2015-01-07
usb/device-info: Listen for USB device add and remove events.

reillyeon:
This patch is an example of using the chrome.usb.onDeviceAdded and
chrome.usb.onDeviceRemoved events to maintain a list of connected
devices.

PR #363, opened 2015-02-17
[usb/device-info] Listen for USB device add and remove events

beaufortfrancois:
This patch is an updated copy of #347.


beaufortfrancois:
Updated webstore at https://chrome.google.com/webstore/detail/usb-device-info/igkmggljimacfdfalpeelenjeicmfnll

PR #367, opened 2015-02-26
Update buy.js and README file to point to new in-app payments library.

mkruisselbrink:


beaufortfrancois:
Hello Marijn, I noticed the "buy.js" file changed. 
Do we need to update https://developer.chrome.com/webstore/payments-iap as well to document the new `google.payments.inapp.consumePurchase`?

jakeleichtling:
Yes, we do. Thanks for pointing this out. We'll make sure our documentation team is aware.

Meggin:
I updated the payments-iap doc src yesterday-- and I just tested a live
link and it looks good. Do you want to test again?

I still need to update the reference though-- so thanks for the heads up.

Meggin

jakeleichtling:
The live page looks good (except for the sample app link, which I've just sent a quick email about). I had posted the above comment before testing. Thanks for making the changes!

PR #375, opened 2015-03-30
[webview browser] Added clear data button

beaufortfrancois:
Since we now have a proper clear data functionality in `<webview>`, let's add it to our webview browser sample.

R=@lazyboy


beaufortfrancois:
Browser sample has been updated in the Chrome Web Store as well at https://chrome.google.com/webstore/detail/edggnmnajhcbhlnpjnogkjpghaikidaa

Issue #371, opened 2015-03-12
contextMenus sample is subject to race conditions


Rob--W:
When two app windows (A and B) are opened, there is a possibility that the API calls interleave. A and B share the same context menu logic (in menus.js), and the following situation can occur:

1. Window A, onload: call chrome.contextMenus.remove
2. Window B, onload: call chrome.contextMenus.remove
3. Window A, callback of contextMenus.remove: Add context menu with a fixed ID.
4. Window B, callback of contextMenus.remove: Add context menu with a fixed ID (ERROR, IDs must be unique).

The code expects / assumes contextmenus.remove + contextMenus.create to be an atomic operation, which is clearly not true given the previous counter-example. A way to solve the problem is to create the common context menu entries only once, and explicitly specify which items have to be removed.

(I discovered this issue after a tester reported a "regression" in https://code.google.com/p/chromium/issues/detail?id=466478)

Another bug is that it is also possible that context menu entries do not appear, because the context menus are removed+created again upon focus of the window. To solve this problem, you could initialize the context menu only once, and restrict the context menu to a specific (app) window via the `documentUrlPatterns` key.

beaufortfrancois:
Thanks Rob! 
I will have a look ;)

beaufortfrancois:
And we are good hopefully!
Thanks again for reporting this.

PR #376, opened 2015-03-30
[context-menus] Use `documentUrlPatterns` to target windows

beaufortfrancois:
As @rob--w mentioned it, this sample was failing to properly create context menus. This patch hopefully addresses this issue. 

BUG=#371
R=@krockot


krockot:
LGTM

PR #377, opened 2015-03-31
Updated camera-capture sample to support multiple sources

lostsource:
- If multiple cameras are detected user can choose source via drop-down
- Camera can now be enabled and disabled

Screenshot of changed UI here: http://imgur.com/MSbXjCO

### DIFF
--- a/samples/camera-capture/app.js
+++ b/samples/camera-capture/app.js
@@ -23,10 +23,46 @@ Author: Eric Bidelman (ericbidelman@chromium.org)
  *
  * @see http://developer.chrome.com/apps/manifest.html#permissions
  */
+
+var curStream = null; // keep track of current stream
+
 function getCamera() {
-  navigator.webkitGetUserMedia({audio: true, video: true}, function(stream) {
-    document.querySelector('video').src = URL.createObjectURL(stream);
+  var cameraSrcId = document.querySelector('select').value;
+
+  // constraints allow us to select a specific video source 
+  var constraints = {
+    video: {
+      optional: [{
+        sourceId: cameraSrcId
+      }]
+    },
+    audio:true
+  }
+
+  // stop any previous stream (eg. while switching between cameras)
+  if(curStream) {
+    curStream.stop();
+  }
+
+  navigator.webkitGetUserMedia(constraints, function(stream) {
+    var videoElm = document.querySelector('video');
+
+    videoElm.src = URL.createObjectURL(stream);
+    curStream = stream;
+
+    curStream.onended = function() {
+      updateButtonState();
+      videoElm.src = "";
+    }
+
+    videoElm.onplay = function() {
+      updateButtonState();  
+    }
+
+    videoElm.play();
+
   }, function(e) {
+    curStream = null;
     console.error(e);
   });
 }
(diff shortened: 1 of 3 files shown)


samples/camera-capture/app.js
```diff
@@ -23,10 +23,46 @@ Author: Eric Bidelman (ericbidelman@chromium.org)
...
+    curStream.onended = function() {
+      updateButtonState();
+      videoElm.src = "";
+    }
+
+    videoElm.onplay = function() {
+      updateButtonState();  
+    }
+
+    videoElm.play();
```
beaufortfrancois: Since `<video>` element is in `autoplay` already, I don't think we need this line there.

samples/camera-capture/app.js
```diff
@@ -23,10 +23,46 @@ Author: Eric Bidelman (ericbidelman@chromium.org)
...
+  var cameraSrcId = document.querySelector('select').value;
+
+  // constraints allow us to select a specific video source 
+  var constraints = {
+    video: {
+      optional: [{
+        sourceId: cameraSrcId
+      }]
+    },
+    audio:true
```
beaufortfrancois: Even though it was in the constraints before, we may want to set `audio` to `false` since we don't need it. What do you think?

samples/camera-capture/app.js
```diff
@@ -35,5 +71,80 @@ function getCamera() {
...
+  else {
+    getCamera();
+  }
+
 });
+
+/**
+ * Change stream source according to dropdown selection
+ */
+document.querySelector('select').onchange = function() {
```
beaufortfrancois: How about `addEventListener('change',` instead of `onchange`?

samples/camera-capture/app.js
lostsource: Wouldn't be a bad idea as the first time I launched the sample I got a lot of annoying feedback sound..

samples/camera-capture/app.js
```diff
@@ -35,5 +71,80 @@ function getCamera() {
...
+        videoSources.push({
+          id: source.id,
+          label: source.label || 'Camera '+(videoSources.length+1)
+        });  
+      }
+    });
+
+    callback(videoSources);
+  });
+}
```
beaufortfrancois: Do you mind adding a Linux "end of line" at the end of the file?

samples/camera-capture/app.js
```diff
@@ -23,10 +23,46 @@ Author: Eric Bidelman (ericbidelman@chromium.org)
...
+  }
+
+  navigator.webkitGetUserMedia(constraints, function(stream) {
+    var videoElm = document.querySelector('video');
+
+    videoElm.src = URL.createObjectURL(stream);
+    curStream = stream;
+
+    curStream.onended = function() {
+      updateButtonState();
```
beaufortfrancois: Each time we switch camera, button label change from "Disable" to "Enable" to "Disable". 
Is there a way to prevent this?

samples/camera-capture/app.js
```diff
@@ -23,10 +23,46 @@ Author: Eric Bidelman (ericbidelman@chromium.org)
...
+    },
+    audio:true
+  }
+
+  // stop any previous stream (eg. while switching between cameras)
+  if(curStream) {
+    curStream.stop();
+  }
+
+  navigator.webkitGetUserMedia(constraints, function(stream) {
```
beaufortfrancois: It would be nice to add something like: `
navigator.getUserMedia = navigator.getUserMedia || navigator.webkitGetUserMedia;` before.

beaufortfrancois:
Thank you so much for this PR.
I've added some comments to it.

beaufortfrancois:
Yes, it would be easier to have one branch per GitHub PR.
No problem for the question. Thank *you* for pushing these changes.

lostsource:
This pull request can be closed as all changes are now in pull request https://github.com/GoogleChrome/chrome-app-samples/pull/378 

I created a separate branch multiCamera as to not mix up changes to other samples

PR #383, opened 2015-04-02
[udp] Updated sample to use the newer chrome.sockets.udp API

lostsource:
Thought this sample should be updated since chrome.socket API has been deprecated for quite a while.

### DIFF
--- a/samples/udp/networking.js
+++ b/samples/udp/networking.js
@@ -4,75 +4,69 @@
     var NoAddressException = "No Address";
     var NotConnectedException = "Not Connected";
 
-    var socket = chrome.socket;
+    var socket = chrome.sockets.udp;
 
-    var baseClient = function(socketMode) {
-      var address;
+    var baseClient = function() {
+      var address, port;
       var socketInfo;
       var connected = false;
       var callbacks = [];
       var self = this;
 
-      this.connect = function(inAddress, port, callback, responseHandler) {
+      this.connect = function(inAddress, inPort, callback, responseHandler) {
         if(!!inAddress == false) throw NoAddressException;
 
         address = inAddress;
-        port = port || this.defaultPort;
-        console.debug('creating socket', socketMode, address, port);
-        socket.create(socketMode, {}, function(_socketInfo) {
+        port = inPort || this.defaultPort;
+
+        console.debug('creating socket', address, port);
+
+        socket.create({}, function(_socketInfo) {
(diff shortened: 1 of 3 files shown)


beaufortfrancois:
Will have a look at it Tuesday for sure ;)

samples/udp/networking.js
```diff
@@ -4,54 +4,47 @@
...
             connected = (connectResult == 0);
-            socket.ondata = function(result) {
+
+            chrome.sockets.udp.onReceive.addListener(function(result){
               if (callbacks.length > 0) {
                 callbacks.shift()(result);           
-              }
-            };
-            self.poll();
+              }
```
beaufortfrancois: Do you mind removing trailing whitespaces there?

samples/udp/networking.js
```diff
@@ -4,54 +4,47 @@
...
+        socket.create({}, function(_socketInfo) {
           socketInfo = _socketInfo;
-          socket.connect(socketInfo.socketId, address, port, function(connectResult) {
-            console.debug('connectResult', connectResult);
+
+          socket.bind(socketInfo.socketId, address, 0, function(connectResult) {
             connected = (connectResult == 0);
-            socket.ondata = function(result) {
+
+            chrome.sockets.udp.onReceive.addListener(function(result){
```
beaufortfrancois: `s/chrome.sockets.udp/socket/`

samples/udp/networking.js
```diff
@@ -107,7 +102,7 @@
       // Clients
       clients: {
         udp: {
```
beaufortfrancois: Since we don't use other types of clients, we may want to remove the "udp" key and simply call `var ec = new chromeNetworking.clients.echoClient();` in demo.js. What do you think?

beaufortfrancois:
Thanks Joseph! Small nits there!
Moreover as in https://github.com/GoogleChrome/chrome-app-samples/pull/385, can you add a space between function params?

lostsource:
Hi Francois, 

I updated this branch as per your comments above.. however I'm not sure if I made the right thing here. Basically I re-based and squashed several commits into commit e3d1fda and push (forced) to update this PR.

Would this be the right approach to squash the 14 commits in PR 385 to a single commit ?

beaufortfrancois:
That's perfect Joseph!
Thank you really much for your work there!

PR #385, opened 2015-04-03
Webserver sample updated to use chrome.sockets.tcpServer

lostsource:
This should help resolve GoogleChrome/chrome-app-samples/issues/384

### DIFF
--- a/samples/webserver/index.js
+++ b/samples/webserver/index.js
@@ -174,7 +174,7 @@ onload = function() {
     var uriEnd = data.indexOf(" ", 4);
     if (uriEnd < 0) { /* throw a wobbler */ return; }
     var uri = data.substring(4, uriEnd);
-    // strip qyery string
+    // strip query string
     var q = uri.indexOf("?");
     if (q != -1) {
       uri = uri.substring(0, q);
--- a/samples/webserver/index.js
+++ b/samples/webserver/index.js
@@ -5,14 +5,16 @@ onload = function() {
   var port = document.getElementById("port");
   var directory = document.getElementById("directory");
 
-  var socket = chrome.socket;
-  var socketInfo;
+  var tcpServer = chrome.sockets.tcpServer;
+  var tcpSocket = chrome.sockets.tcp;
+
+  var serverSocketId = null;
   var filesMap = {};
 
   var stringToUint8Array = function(string) {
     var buffer = new ArrayBuffer(string.length);
     var view = new Uint8Array(buffer);
-    for(var i = 0; i < string.length; i++) {
+    for (var i = 0; i < string.length; i++) {
       view[i] = string.charCodeAt(i);
     }
     return view;
@@ -21,7 +23,7 @@ onload = function() {
   var arrayBufferToString = function(buffer) {
     var str = '';
     var uArrayVal = new Uint8Array(buffer);
-    for(var s = 0; s < uArrayVal.length; s++) {
+    for (var s = 0; s < uArrayVal.length; s++) {
       str += String.fromCharCode(uArrayVal[s]);
     }
     return str;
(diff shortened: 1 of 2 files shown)


samples/webserver/index.js
```diff
@@ -67,65 +100,61 @@ onload = function() {
...
-      }
-      else {
-        // Throw an error
-        socket.destroy(socketId);
-      }
-    });
+    if (acceptInfo.socketId != serverSocketId)
+      return; 
+
+    console.warn("ACCEPT", acceptInfo)
```
beaufortfrancois: Why not `console.log` there?

samples/webserver/index.js
```diff
@@ -43,21 +81,16 @@ onload = function() {
...
-    console.info("writeErrorResponse::filereader:: end onload...");
 
+    sendReplyToSocket(socketId, outputBuffer,keepAlive);
+
+    console.info("writeErrorResponse::filereader:: end onload...");
     console.info("writeErrorResponse:: end...");
   };
 
   var write200Response = function(socketId, file, keepAlive) {
+    console.log("write200Response",socketId,keepAlive);
```
beaufortfrancois: Do we need this log there?

samples/webserver/index.js
```diff
@@ -29,6 +31,42 @@ onload = function() {
...
+    }
+
+    tcpServer.onAccept.removeListener(onAccept);
+    tcpSocket.onReceive.removeListener(onReceive);
+  }
+
+  var sendReplyToSocket = function(socketId,buffer,keepAlive) {
+    // verify that socket is still connected before trying to send data
+    tcpSocket.getInfo(socketId,function(socketInfo) {
+      if(socketInfo.connected !== true) {
```
beaufortfrancois: `if (!socketInfo.connected) {` seems easier to read than `if(socketInfo.connected !== true) {`

samples/webserver/index.js
```diff
@@ -29,6 +31,42 @@ onload = function() {
...
+        destroySocketById(socketId);
+        return;
+      }
+
+      tcpSocket.setKeepAlive(socketId, keepAlive, function() {
+        tcpSocket.send(socketId, buffer, function(writeInfo) {
```
beaufortfrancois: You may want to check `chrome.runtime.lastError` there also since I happen to have an error on Chrome OS Dev Channel. For info, setting a delay to `1` fixed the raised error.

For info, here's the `ui.LATEST` error I get when running it with no delay:
`[31544:31565:0408/094629:ERROR:tcp_socket_libevent.cc(72)] Failed to set TCP_KEEPIDLE on fd: 704: Invalid argument`
Hopefully @reillyeon will be able to help us there.

samples/webserver/index.js
```diff
@@ -29,6 +31,42 @@ onload = function() {
...
+
+  var sendReplyToSocket = function(socketId,buffer,keepAlive) {
+    // verify that socket is still connected before trying to send data
+    tcpSocket.getInfo(socketId,function(socketInfo) {
+      if(socketInfo.connected !== true) {
+        destroySocketById(socketId);
+        return;
+      }
+
+      tcpSocket.setKeepAlive(socketId, keepAlive, function() {
```
beaufortfrancois: You may want to check `chrome.runtime.lastError` there also since I happen to have an error on Chrome OS Dev Channel. For info, setting a delay to `1` fixed the raised error.

For info, here's the `ui.LATEST` error I get when running it with no delay:
`[31544:31565:0408/094629:ERROR:tcp_socket_libevent.cc(72)] Failed to set TCP_KEEPIDLE on fd: 704: Invalid argument`
Hopefully @reillyeon will be able to help us there.

samples/webserver/index.js
```diff
@@ -29,6 +31,42 @@ onload = function() {
...
+        tcpSocket.send(socketId, buffer, function(writeInfo) {
+          console.log("WRITE", writeInfo,"socketId",socketId);
+
+          if (keepAlive !== true || chrome.runtime.lastError) {
```
beaufortfrancois: How about `!keepAlive` instead of `keepAlive !== true`?

beaufortfrancois:
As usual, thank you for doing this!
Here are some general feedback on top of the comments above:
- Can you add a space between function params? 
`tcpSocket.getInfo(socketId,function(socketInfo) {` becomes `tcpSocket.getInfo(socketId, function (socketInfo) {`
- It would be nice to have only one commit. Right now, it looks we have 14 according to GitHub.

samples/webserver/index.js
lostsource: I set the delay to 1 and I'm checking for `chrome.runtime.lastError`, if an error occurs I'm aborting the `send` and closing the socket. Not sure if this is the right way to do it, or we could still call `send` after `setKeepAlive` fails. (couldn't reproduce on Canary to properly test this)

lostsource:
Branch updated as per PR comments and rebased to 1 commit.

I re-wrote the way response headers were being set as the line was a tad too long, so headers are now handled by the getResponseHeader function.

samples/webserver/index.js
```diff
@@ -29,113 +31,177 @@ onload = function() {
...
+
+    var keepAlive = false;
+    if (data.indexOf("Connection: keep-alive") != -1) {
+      keepAlive = true;
+    }
+
+    var uriEnd = data.indexOf(" ", 4);
+    if (uriEnd < 0) { /* throw a wobbler */ return; }
+    var uri = data.substring(4, uriEnd);
+    // strip qyery string
```
beaufortfrancois: Little nit I've missed: `s/qyery/query`

samples/webserver/index.js
```diff
@@ -145,11 +211,15 @@ onload = function() {
...
 
   start.onclick = function() {
-    socket.create("tcp", {}, function(_socketInfo) {
-      socketInfo = _socketInfo;
-      socket.listen(socketInfo.socketId, hosts.value, parseInt(port.value), 50, function(result) {
+
+    tcpServer.create({}, function(socketInfo) {
+      serverSocketId = socketInfo.socketId;
+
+      tcpServer.listen(serverSocketId, hosts.value, parseInt(port.value, 10), 50, function(result) {
```
beaufortfrancois: I appreciate the base 10 for `parseInt` there ;)

beaufortfrancois:
Can we bump app version in a subsequent commit as part of this PR?
Otherwise :+1: !

lostsource:
PR updated as per comments

beaufortfrancois:
Perfect!

image

beaufortfrancois:
Webstore Item has been updated https://chrome.google.com/webstore/detail/http-server-sample/hflhcpmgeolmjlbmdicgkjedjmkoocbe

PR #391, opened 2015-04-30
merge Tapted master

tapted:


tapted:
whoops wrong upstream

Issue #396, opened 2015-06-10
Update chrome.hid sample to use chrome.hid.getUserSelectedDevices API


reillyeon:
Updating the HID sample app to use this new API (about to land in dev channel) will allow developers to try out the HID API with their devices without having to modify the app manifest to include permission for their device.

beaufortfrancois:
If that's ok with you, I'll submit a PR since https://codereview.chromium.org/1115213004 has been accepted and I'll be glad if you can review it.

reillyeon:
Go right ahead!

beaufortfrancois:
I'll close this issue since all PR have been merged.

PR #404, opened 2015-06-25
[hid] Add onDeviceAdded and onDeviceRemoved events

beaufortfrancois:
Hey @reillyeon,

Before adding `chrome.hid.getUserSelectedDevices`, I thought it was better to add `onDeviceAdded` and `onDeviceRemoved` event listeners to the `hid` sample instead of polling.

BUG=#396

### DIFF
--- a/samples/hid/control-panel.js
+++ b/samples/hid/control-panel.js
@@ -16,8 +16,6 @@
   };
 
   var connection = -1;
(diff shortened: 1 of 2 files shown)


beaufortfrancois:
Thank you @reillyeon for the feedback.
I've removed the `filters` parameter.

PR #408, opened 2015-06-29
[hid] Add chrome.hid.getUserSelectedDevices

beaufortfrancois:
Hey @reillyeon,

This CL simply adds the "Add device" button with `chrome.hid.getUserSelectedDevices`. I've also removed the "usb_devices" permissions from the manifest file since I think they don't matter anymore.

BUG=#396


beaufortfrancois:
I've set `minimum_chrome_version` to `45.0.2439.3` and removed check for `chrome.hid.getUserSelectedDevices`

PR #412, opened 2015-08-12
[webview browser] Fix find-box and zoom-box UI alignments

oahziur:
This is related to https://code.google.com/p/chromium/issues/detail?id=519440.


beaufortfrancois:
Thank you!

PR #429, opened 2016-02-18
Update websocket-server to use chrome.sockets API.

rhashimoto:
The chrome.socket API is deprecated since Chrome 33. This commit
updates websocket-server to the chrome.sockets.tcp and
chrome.sockets.tcpServer APIs.

For convenience, the chrome.sockets.tcp socketId is wrapped with a
PSocket class with member functions read/write/close that return
Promises.

### DIFF
--- a/samples/websocket-server/http.js
+++ b/samples/websocket-server/http.js
@@ -349,7 +349,15 @@ HttpServer.prototype = {
         port,
         50,
         function(result) {
-          t.readyState_ = 1;
+          if (!result) {
+            t.readyState_ = 1;
+          }
+          else {
+            console.log(
+              'listen error ' +
+              chrome.runtime.lastError.message +
+                ' (normal if another instance is already serving requests)');
+          }
         });
     });
   },
--- a/samples/websocket-server/http.js
+++ b/samples/websocket-server/http.js
@@ -663,6 +663,7 @@ WebSocketRequest.prototype = {
  */
 function WebSocketServerSocket(pSocket) {
   this.pSocket_ = pSocket;
+  this.readyState = 1;
   EventSource.apply(this);
   this.readFromSocket_();
 }
@@ -690,8 +691,10 @@ WebSocketServerSocket.prototype = {
    * process.
    */
   close: function() {
-    this.sendFrame_(8);
-    this.readyState = 2;
+    if (this.readyState === 1) {
+      this.sendFrame_(8);
+      this.readyState = 2;
+    }
   },
 
   readFromSocket_: function() {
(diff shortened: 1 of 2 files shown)


samples/websocket-server/http.js
```diff
@@ -252,46 +337,34 @@ HttpServer.prototype = {
...
-                    function(result) {
-        t.readyState_ = 1;
-        t.acceptConnection_(t.socketInfo_.socketId);
+    chrome.sockets.tcpServer.create(function(socketInfo) {
+      chrome.sockets.tcpServer.onAccept.addListener(function(acceptInfo) {
+        if (acceptInfo.socketId === socketInfo.socketId)
+          t.readRequestFromSocket_(new PSocket(acceptInfo.clientSocketId));
       });
+      
+      chrome.sockets.tcpServer.listen(
```
beaufortfrancois: When running this, I get this error below.
Please consider checking runtime.lastError. I get the "net::ERR_FILE_NOT_FOUND" error message as you can see.
```
extensions::lastError:133 Unchecked runtime.lastError while running sockets.tcpServer.listen: net::ERR_FILE_NOT_FOUND
    at Object.callback (chrome-extension://klacjkbkhfacpdolfbaehocfpjjipffa/http.js:346:32)

File: samples/websocket-server/manifest.json [hunk 2]
```diff
@@ -4,10 +4,17 @@
...
+    "tcp": {
+      "connect": "*"
+    },
+
+    "tcpServer": {
+      "listen": "*"
+    }
+  },
+  
+  "permissions": [],
```
beaufortfrancois: I think we can omit the "permissions" key there.

beaufortfrancois:
Thank you Roy for doing this.
Except two nits, it looks good to me.

samples/websocket-server/http.js
rhashimoto: Technically this is not a regression because the trunk version also does not handle `lastError` :-). But I fixed it.

rhashimoto:
Oh, please note that I did *not* change the association with `old_sockets` in the tables in the samples root `README.md`. I wasn't sure if I should be doing that or not.

beaufortfrancois:
Thank you Ryo for taking care of these (old) nits.
I'll merge this CL and will update README file in a minute.

beaufortfrancois:
Here it is: https://github.com/GoogleChrome/chrome-app-samples/compare/7fde41f807860bdab90046f217732dbb8f26f03b...e347c538e8612aa3b0f90bde0fc721c4f0569125
And the sample has been updated as well at https://chrome.google.com/webstore/detail/pkbpddppnkjmlbgliipgmhjeialadokj

PR #474, opened 2017-10-10
Added De-registration to gcm-notifications.

nzcoffeem:


petewil-G:
The change looks good, but the API is no longer supported in Chrome, so there isn't much point in taking the change, let me know if you disagree.  The original gcm-notifications was replaced by "gcm for chrome" about 2 years ago, and this API has been turned off everywhere.

petewil-G:
Thanks for making the change, by the way!  I appreciate that you spent time on it, and I'm sorry that the time has gone to waste.  I also appreciate being corrected.

nzcoffeem:
No worries. I was able to use the gcm-notifications chrome app though so the API has not been turned off and still connects to the GCM push network. The tool was good for testing push notifications to GCM to simulate an Android App 👍 .

PR #542, opened 2021-01-22
Move MV2 resources into dedicated archive directory

dotproto:
This PR paves the way for a new set of MV3 resources that will be the default set of resources moving forward. We're not ready to drop MV2 resources, but we also don't want them to be the main content everyone comes here for.

This migration approach aims to make updating MV2 resource urls relatively straightforward: just add 'mv2-archive/' to the beginning of existing paths.

Apps remain a top level directory because they are going through their own deprecation cycle, but they could potentially be moved into the mv2 directory as well since the app platform is built on top of MV2 extensions.


awfuchs (approved):

rvong:
This breaks all the links in the README.md

dotproto:
@rvong whoops! Thanks for the heads up, fixing now.

Pakinay (commented):
mv2-archive/api/bookmarks/basic/icon.png

Pakinay (commented):
mv2-archive/tmv2-archive/api/browserAction/print/manifest.jsonutorials/oauth_tutorial_complete/oauth.jsmv2-archive/api/browserAction/print/background.js

PR #552, opened 2021-02-18
Cookie Clearer example

dotproto:
Adds a MV3 sample for clearing all cookies on a given domain. This is an adaptation of a demo originally written by ghazale on the Chrome Extensions team.


api/cookies/cookie_clearer/popup.js
```diff
@@ -0,0 +1,101 @@
+const form = document.getElementById("control-row");
+const go = document.getElementById("go");
+const input = document.getElementById("input");
+const message = document.getElementById("message");
+
+
+document.addEventListener("DOMContentLoaded", initPopupWindow);
```
samthor, line 7: Why do this? Your script runs synchronously where it's loaded; all the DOM you care about is above this.

I'd rather you use `type="module"` (which is implicit defer) and then use top-level `await` for the setup. (Chrome's samples regularly have `async` handlers for events, which .... I get, we have async APIs, but is hard).

api/cookies/cookie_clearer/popup.js
```diff
@@ -0,0 +1,101 @@
...
+form.addEventListener("submit", handleFormSubmit);
+
+async function initPopupWindow() {
+  let [tab] = await chrome.tabs.query({ active: true, currentWindow: true });
+
+  if (tab.url) {
```
samthor, line 13: Is it possible that `tab` is undefined here?

README.md
```diff
@@ -20,3 +20,29 @@ The directory structure is as follows:
...
+          <li><a href="https://developer.chrome.com/docs/extensions/reference/cookies/#method-getAll">cookies.getAll</a></li>
+          <li><a href="https://developer.chrome.com/docs/extensions/reference/cookies/#method-remove">cookies.remove</a></li>
+          <li><a href="https://developer.chrome.com/docs/extensions/reference/tabs/#method-query">tabs.query</a></li>
+        </ul>
+      </td>
+    </tr>
+  </tbody>
+</table>
+
+Manifest V2 extensions can be found
```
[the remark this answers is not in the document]
dotproto: Whoops, shouldn't have included that – there already mentioned a bit further up in the page. Removing that reference now.

api/cookies/cookie_clearer/popup.js
dotproto: It is technically possible, yes. I'll add tab to the if check to make sure it's not undefined.

api/cookies/cookie_clearer/popup.js
dotproto: This is an artifact of the original implementation. The page originally inserted popup.js in the head & used DOMContentLoaded to defer execution.

I actually hadn't considered using top level await. This is largely personal preference, but one of the things I like about leaning on DOMContentLoaded is that it clearly indicates when that initialization logic is executed. Without it, that timing is implicit based on attributes set outside the script's control (DOM order, script attributes, etc.).

According to Chrome Status, [top level await](https://www.chromestatus.com/feature/5767881411264512) isn't available until M89. For the moment I've wrapped it in an async IIFE.

README.md
```diff
@@ -32,7 +32,7 @@ Read more on [Getting Started](https://developer.chrome.com/extensions/getstarte
     <tr>
       <td style="vertical-align:top;">
         Cookie Clearer <br>
-        <a href="api/cookies/cookie_clearer"><code>api/cookies/cookie_clearer</code></a>
+        <a href="api/cookies/cookie-clearer"><code>api/cookies/cookie_clearer</code></a>
```
samthor, line 35: should you update the visible text to match the link?

api/cookies/cookie_clearer/popup.js
samthor: `DOMContentLoaded` isn't as useful as you think: https://whistlr.info/2020/understanding-load/#use-cases

Thanks for removing it though. Just running the code is clearer.

api/cookies/cookie-clearer/popup.js
```diff
@@ -0,0 +1,100 @@
+const form = document.getElementById("control-row");
+const go = document.getElementById("go");
+const input = document.getElementById("input");
+const message = document.getElementById("message");
+
+// The async IIFE is necessary because Chrome <89 does not support top level await.
+(async function initPopupWindow() {
+  let [tab] = await chrome.tabs.query({ active: true, currentWindow: true });
+
+  if (tab && tab.url) {
```
samthor, line 10: Not required, but you could also do `if (tab?.url) {` now.

samthor (approved):

api/cookies/cookie-clearer/popup.js
dotproto: "Optional Chaining reached Stage 3 at the July 2019 TC39 meeting." [Chrome Status](https://www.chromestatus.com/feature/5748330720133120)

I'll update to include this.

README.md
dotproto: D'oh. Fixed!

PR #553, opened 2021-02-18
Adds simple Hello World example

dotproto:
I'm trying something a little different here – explaining some of the basic concepts in inline comments. I'm curious if y'all have any thoughts about the execution here.


README.md
```diff
@@ -14,9 +14,33 @@ The directory structure is as follows:
...
+  </thead>
+  <tbody>
+    <tr>
+      <td style="vertical-align:top;">
+        Hello World <br>
+        <a href="examples/hello-world"><code>examples/hello-world</code></a>
+      </td>
+      <td  style="vertical-align:top;">
+        <ul>
+          <li><a href="https://developer.chrome.com/docs/extensions/reference/action/#event-onClicked">cookies.getAll</a></li>
```
samthor, line 39: This first link seems wrong?

examples/hello-world/background.js
```diff
@@ -0,0 +1,21 @@
+// Extension event listeners are a little different from the patterns you may hav seen in DOM or
```
samthor, line 1: hav => have

examples/hello-world/background.js
```diff
@@ -0,0 +1,21 @@
...
+// Node.js APIs. Our event listener registration can be broken in to 4 distinct parts:
+//
+// * chrome      - the global namespace for extension APIs.
+// * action      – the namespace of the specific API we want to use.
+// * onClicked   - the event we want to subscribe to.
+// * addListener - what we want to do with this event.
```
samthor, line 7: I'd say "the method which adds the handler" here.

examples/hello-world/background.js
```diff
@@ -0,0 +1,21 @@
...
+
+  // While we could have used `let url = "hello.html"`, using runtime.getURL is a bit more robust as
+  // it gives you a full URL rather than just a path that Chrome must needs to be resolve
+  // contextually at runtime.
+  let url = chrome.runtime.getURL("hello.html");
+
+  // Create a new tab and point it at our page's URL using JavaScript's object initializer
+  // shorthand. For more information on this syntax see
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Object_initializer#new_notations_in_ecmascript_2015
+  chrome.tabs.create({ url });
```
samthor, line 20: Should you demonstrate the callback or `Promise` version here, by adding either of those things?

examples/hello-world/background.js
dotproto: The intent of the current copy was to focus on the role of the last parameter. While we are registering the event listener, we could call a number of other methods in this position. The full list of methods is available at https://developer.chrome.com/docs/extensions/reference/events/#type-Event.

Any thoughts on how we can better communicate this intent?

examples/hello-world/background.js
dotproto: Just updated the code sample to use async/await, log the ID of the tab we created, and add guidance on how to open devtools.

README.md
dotproto: Yep yep, bad copy paste. Thanks for catching this.

examples/hello-world/background.js
dotproto: Why oh why didn't my spell checker catch that? 😐

dotproto:
Updated the sample to open a new tab on installation rather than requiring the user to click the action. I think this better accomplishes the goal of providing immediate feedback on installation, but comes at the cost of not leveraging a common pattern that we want to encourage (responding to action invocation).

examples/hello-world/background.js
samthor: I see your point. I think for the vast majority of people this is all they use so it feels a bit superfluous. But I'm not going to argue this point, I think it's fine.

samthor (approved):

peteos123 (commented):
Hello World!

PR #567, opened 2021-03-10
Port bookmarks example to MV3

dotproto:
This is a naive port of the MV2 bookmarks example to MV3. It updates the existing example to work with MV3, but does not meaningfully modernize the implementation or demonstrate best practices. When feasible, this example should be replaced.


Harper9 (approved):

api/bookmarks/popup.js
```diff
@@ -0,0 +1,141 @@
...
+    list.append(dumpNode(bookmarkNodes[i], query));
+  }
+
+  return list;
+}
+
+function dumpNode(bookmarkNode, query) {
+  if (bookmarkNode.title) {
+    if (query && !bookmarkNode.children) {
+      if (String(bookmarkNode.title).indexOf(query) == -1) {
```
samthor, line 30: in testing this, I found it being case-sensitive a bit odd. Can you lower-case both before you check?

samthor (approved):
I know you're just porting this code—written with jQuery UI—but it's not particularly good. The dialog overlays don't work so they look like this:
Screen Shot 2021-04-08 at 11 13 48

It's also possible to leave parts of dialogs on the screen even after they're opened.

Having said that, I wouldn't stop you submitting this since it is just a migration, but it definitely needs work.

api/bookmarks/popup.js
dotproto: Done.

dotproto:
> I know you're just porting this code—written with jQuery UI—but it's not particularly good.

I agree. To be honest I'm inclined to drop this PR and replace it with either with a demo more like the Action API one or a ground up rewrite. What do you think?

dotproto:
I had a few offline conversations with folks about how to handle this PR. While it's not a best in class example, ultimately I think it does more good to have a working MV3 version that's a little creaky than to not have a reference at all. I'm going to do some additional cleanup to this PR to address the UX issues @samthor raised, but after that I think we'll be good to merge.

dotproto:
Made some tweaks to the UI in order to make the experience a little better. Dialog windows now appear in the area you initiated the interaction, and there are no longer UI artifacts left at the bottom of the page.

There is a bit of a bug in that when you scroll and open a dialog, for some reason the viewport position is set to the top of the window, then when the animation finishes it pops back to the correct offset. I have no idea what's happening there and at the moment it's not worth investing the time to fix.

With this change the demo is in better shape than it's been in for years.

PR #569, opened 2021-03-11
Port Omnibox suggestion demo to MV3

dotproto:
This PR is blocked by a bug [crbug.com/1186804](https://crbug.com/1186804); I don't think it makes sense to merge an example that does not work in Chrome Stable.

Additionally, this demo may be overly simplistic. I'd like to explore something more in line with the Action API example and try to provide a demo that flexes the various capabilities of this API.


api/omnibox/simple-example/background.js
```diff
@@ -0,0 +1,21 @@
...
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+// This event is fired each time the user updates the text in the omnibox,
+// as long as the extension's keyword mode is still active.
+chrome.omnibox.onInputChanged.addListener((text, suggest) => {
+  console.log('inputChanged: ' + text);
+  suggest([
```
zamiang, line 11: I assume you are working on this since it is an open PR but this line triggers an error when I run this example

```
Error in event handler: ReferenceError: DOMParser is not defined
    at parseOmniboxDescription (extensions::omnibox:24:19)
    at extensions::omnibox:101:25
    at suggestCallback (extensions::omnibox:116:20)
    at chrome-extension://eieflmjhclgndndbllffkimmokpfhibl/background.js:11:3
    at extensions::omnibox:118:3
```

dotproto: Yep, I may end up splitting this PR into 2 as one of the samples works and one does not and I don't want to block on merging this PR until both are functional.

dotproto:
Force pushed to separate the two omnibox examples. This PR is now restricted to the "input suggestions" example.

api/omnibox/input-suggestions/background.js
```diff
@@ -0,0 +1,23 @@
...
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+// This event is fired each time the user updates the text in the omnibox,
+// as long as the extension's keyword mode is still active.
+chrome.omnibox.onInputChanged.addListener((text, suggest) => {
+  console.log('inputChanged: ' + text);
+  suggest([
```
samthor, line 11: I don't see these suggestions appear anywhere.

api/omnibox/input-suggestions/background.js
```diff
@@ -0,0 +1,23 @@
...
+    {content: text + " one", description: "the first one"},
+    {content: text + " number two", description: "the second entry"}
+  ]);
+});
+
+// This event is fired with the user accepts the input in the omnibox.
+chrome.omnibox.onInputEntered.addListener((text) => {
+  console.log('inputEntered: ' + text);
+  chrome.tabs.create({
+    url: `data:text/html;charset=utf-8,You entered "${text}"`
```
samthor, line 21: Sorry to make you expand this, but can you make this a real page with a query-string?

That page would be included in this extension and have a tiny note about "You can open a new tab in response to an omnibox search, but it's not the only possibility". I think it might confuse our users who think the only thing you can do is open a page.

samthor (approved):
This extension works for me in Canary?

api/omnibox/input-suggestions/background.js
samthor: I'd actually like to have the opening behavior specifically look for these entries and do something special based on whichever one you choose.

samthor (changes requested):
Actually I'm un-approving until we can see the suggestions. I'd actually also suggest remove the page-opening behavior since it's shown off in the other demo, and focus entirely on suggestions and what you can do with them.

api/omnibox/input-suggestions/background.js
dotproto: > I don't see these suggestions appear anywhere.

This is caused by a Chrome bug. This API is partially implemented by JS in Chrome and that JS tries to use DOMParser. Since service workers don't have access to DOM related capabilities, this causes Chrome to throw.

I think this bug blocks merging this PR as the example isn't usable in it's current form. I'll add a note about this to the main PR comment. 

> I'd actually like to have the opening behavior specifically look for these entries and do something special based on whichever one you choose.

Off the top of my head one example might be to open a Google search term in the current window, a new tab, or a new window. I'll run with that for now, but if you (or anyone else) have a better idea I'm happy to do something else.

samthor (commented):
Please let me know when the underlying bug is fixed and we can review this :-)

dotproto:
I'm going to close this review as it's not representative of the type of demos I'm currently trying to pursue. I'd like to create a new demo more in line with the Action API example, but I'll have to put that on the back burner until the Omnibox API works as expected in Canary (at the earliest).

PR #574, opened 2021-03-24
Action API demo

dotproto:
This example takes a bit of a different approach demonstrating how extensions APIs work. Rather than target a specific use case, we try to demonstrate at least one way that each method on the API can be used. Feedback on this approach as well as the content of the PR is welcome.


api/action/index.html
```diff
@@ -0,0 +1,109 @@
...
+  <title>Document</title>
+  <script defer src="index.js"></script>
+</head>
+<body>
+  <p>Before experimenting with these APIs, we recommend you pin the extension's action button to your
+  toolbar in order to make it easier to see the changes. </p>
+
+  <section id="toggle-state">
+    <h2>enable / disable</h2>
+    <p>Clicking the button will toggle the action button's enabled/disabled state.</p>
```
samthor, line 16: Nit: it wasn't obvious to me that you meant the button right below _on this page_. I tried to click the extension's icon and wondered why it didn't grey out.

api/action/index.js
```diff
@@ -0,0 +1,120 @@
...
+// Badge background color
+
+document.getElementById('set-badge-background-color-button').addEventListener('click', () => {
+  // To show off this method, we must first make sure that there is a badge value
+
+  chrome.action.setBadgeText({ text: 'hi :)' });
+
+  let color = new Array(4);
+  for (let i = 0; i < color.length; i++) {
+    color[i] = Math.floor(Math.random() * 256);
```
samthor, line 64: Does this actually respect opacity?

api/action/index.js
```diff
@@ -0,0 +1,120 @@
...
+
+let lastIconIndex = 0;
+document.getElementById('set-icon-button').addEventListener('click', async () => {
+  lastIconIndex = (lastIconIndex + 1) % EMOJI.length;
+  let emojiFile = `images/emoji-${EMOJI[lastIconIndex]}.png`;
+
+  let response = await fetch(emojiFile);
+  let blob = await response.blob();
+  let imageBitmap = await createImageBitmap(blob);
+  let osc = new OffscreenCanvas(imageBitmap.width, imageBitmap.height);
```
samthor, line 85: This doesn't need to be an OffscreenCanvas—it's on the main thread, right?

samthor (commented):
I think this is a really cute demo page and I think it works pretty well. My biggest concern though is basically the get/set flow. I get that it matches the API calls themselves but it's awkward versus:

- being able to hit "enter" while focused on the input box
- having "get" work implicitly (even though, yes, two pages open can't _subscribe_ to the current state).

Can we subscribe to state changes? If so I'd probably suggest streamlining the page. Otherwise... I'm not completely sure.

api/action/index.js
dotproto: While true, this specific case was designed to be usable in service workers as well. I'm assuming that developers will copy-paste sample code directly into their extensions, so I wanted to maximize compatibility. Perhaps it would be best to add a comment to that effect?

More generally I've been debating whether or not the demo API interactions should live in the extension's service worker. The main advantage is that this is likely the more general pattern for extension development, but the down side is that this complexity makes the demo harder to understand and that some developers will assume that these APIs can only be used in a background context.

api/action/index.js
dotproto: It doesn't. An alpha value of 0 will set the badge color to the default value, any other value will set the badge to a fully opaque version of the specified color.

https://bugs.chromium.org/p/chromium/issues/detail?id=1184905

dotproto:
> being able to hit "enter" while focused on the input box

This is mostly an artifact of me being lazy and not actually using forms. We can enhance the demo by either adding forms or (where appropriate) immediately applying the change.

> Can we subscribe to state changes? If so I'd probably suggest streamlining the page. Otherwise... I'm not completely sure.

The Action API doesn't have that kind of event that we could subscribe to, no. That said, we could have form submission await the set result, get the new value, and use that to populate a "current value" input field. 

The design of the current implementation was intentionally trying to provide a simple GUI for the API calls. The idea was basically to show a no-frills implementation of how the API works so that developers can get a feel for it. Maybe that's not the right approach, though.

api/action/index.js
samthor: I'm happy with the argument that this code might run anywhere and users won't have to worry. 👍

api/action/index.js
```diff
@@ -24,97 +37,197 @@ document.getElementById('toggle-state-button').addEventListener('click', (_event
...
   let imageData = ctx.getImageData(0, 0, osc.width, osc.height);
 
   chrome.action.setIcon({ imageData });
 });
 
-// Title
+document.getElementById('reset-icon-button').addEventListener('click', () => {
+  let manifest = chrome.runtime.getManifest();
+  chrome.action.setIcon({ path: manifest.action.default_icon });
+})
```
samthor, line 195: missing trailing semicolon

api/action/popups/b.html
```diff
@@ -5,8 +5,28 @@
   <meta http-equiv="X-UA-Compatible" content="IE=edge">
   <meta name="viewport" content="width=device-width, initial-scale=1.0">
   <title>Document</title>
+  <style>
+    .center {
```
samthor, line 9: Should you unify this CSS in an external file? It's the same for many popups.

api/action/index.js
```diff
@@ -1,14 +1,27 @@
-const EMOJI = [
-  'confetti',
-  'suit',
-  'bow',
-  'dog',
-  'skull',
-  'yoyo',
-  'cat',
-];
+/**
```
samthor, line 1: Sorry to rewrite your code but this should look like:

```js
/**
 * @param {number} timeout
 * @param {(event: Event) => void} callback
 * @return {(event: Event) => void}
 */
function debounce(timeout, callback) {
  let timeoutID = 0;
  return (event) => {
    clearTimeout(timeoutID);
    timeoutID = setTimeout(() => callback(event), timeout);
  };
}
```

The convenient thing here is that you can safely call `clearTimeout` on a zero ID, or an ID which has already fired. It's harmless.

samthor: Also [MDN agrees](https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/setTimeout) that it's safe to always call `clearTimeout` on an oldID:

> It is guaranteed that a timeout ID will never be reused by a subsequent call to setTimeout() or setInterval() on the same object (a window or a worker).

TIL you can probably only ever have [MAX_SAFE_INTEGER](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER) calls to `setTimeout`.

api/action/index.js
```diff
@@ -24,97 +37,197 @@ document.getElementById('toggle-state-button').addEventListener('click', (_event
...
-  });
+async function getCurrentPopup() {
+  let popup = await chrome.action.getPopup({});
+  document.getElementById('current-popup-value').value = popup;
+  return popup;
+};
+
+async function showCurrentPage() {
+  let popup = await getCurrentPopup();
+  if (!popup) return;
```
samthor, line 56: super nit: {}'s please

api/action/index.js
samthor: It feels confusing to set the alpha channel to a random value, then? Perhaps you should have a:

```js
let color;
if (Math.random() < 0.5) {
  // show default color
  color = [0, 0, 0, 0];  // zero opacity uses default color
} else {
  color = new Array(3).map(() => Math.floor(Math.random * 255));
  color.push(255);  // if non-zero, this will be fully opaque
}
chrome.action.setBadgeBackgroundColor({ color });
```

api/action/index.js
```diff
@@ -0,0 +1,233 @@
...
+// the `onclicked-button` handler to prevent the user from accidentally registering multiple
+// onClicked listeners.
+chrome.action.onClicked.addListener((tab) => {
+  chrome.tabs.create({ url: 'https://html5zombo.com/' });
+});
+
+document.getElementById('onclicked-button').addEventListener('click', async () => {
+  // Our listener will only receive the action's click event after clear out the popup URL
+  await chrome.action.setPopup({ popup: '' });
+  await showCurrentPage();
```
samthor, line 81: My minor complaint here is that this action makes the `<select>` not really show the correct state any more (since there's default, A, B.). But it's so minor I'm not even that worried.

samthor (approved):

api/action/index.js
dotproto: I like this suggestion. I'll integrate this today.

api/action/popups/b.html
dotproto: Yep, I was just being lazy :)

api/action/index.js
dotproto: I'm glad for the rewrite. Mine was functional but it felt like it was more complicated than the other debug functions I vaguely recalled. I'll integrate this shortly.

api/action/index.js
dotproto: FYI, `new Array(3).map()` does not work as expected because new Array produces a holy array and `map()` skips holes. A couple alternative oneliners are

```js
// generic
new Array(n).fill().map(…);
[...new Array(n)].map(…);
Array.from(new Array(n).values()).map(…);

// specific
[0,0,0].map(…)
```

In this case I think I'll go with the simplest approach and just use a literal.

api/action/index.js
dotproto: There's already other stuff to clean up, so I'll integrate this now.

manfredwedderin (commented):
[]()

PR #578, opened 2021-03-28
Fix spelling

ssoda:


dotproto (approved):
Thanks for the fixes! My spelling is bad :(

PR #579, opened 2021-03-30
Alarms API demo

dotproto:
First pass on an example Alarms API demo extension. This demo allows users to create extensions using various combinations of [alarms.create](https://developer.chrome.com/docs/extensions/reference/alarms/#method-create) and to visualize the existing existing alarms.


dotproto:
I'm going to take a quick shot at a different approach to this demo. While what I have works, I lost the plot a bit and ended up focusing too much on trying to make something app-like rather than demonstrative. As a result, the current example is harder understand then it should be.

dotproto:
Closing as #580 feels like the better approach.

PR #580, opened 2021-03-30
Alternate Alarms API demo

dotproto:
I wasn't satisfied with the approach I took to demoing the Alarms API in PR #579, so I plugged away at it some more in my free time and came up with a demo that works and looks very different. Feedback welcome :)


api/alarms/background.js
```diff
@@ -0,0 +1,11 @@
+// Open the extension's demo page on install/update
+chrome.runtime.onInstalled.addListener((_reason) => {
+  chrome.tabs.create({
+    url: chrome.runtime.getURL('index.html')
```
samthor, line 4: I'm a bit confused about how someone gets back to this page a 2nd time. Should there be an action to open the page? Can the link be included on the extensions' settings page?

api/alarms/index.html
```diff
@@ -0,0 +1,66 @@
...
+      <div class="create-alarm__value">
+        <input type="number" step="0.1" name="time-value" min="0" value=1>
+
+        <select name="time-format">
+          <option id="format-minutes" value="min" selected>minutes</option>
+          <option id="format-ms" value="ms">milliseconds</option>
+        </select>
+      </div>
+
+      <div class="create-alarm__label">Alarm period *</div>
```
samthor, line 30: I know it's called `alarmPeriod` in the config but this could be better described as "Alarm repeat period".

api/alarms/index.html
```diff
@@ -0,0 +1,66 @@
...
+      <div class="create-alarm__value">
+        <input type="number" step="0.1" min="0" name="period" value="0"> minutes
+        <br><i>Non-zero values create a periodic alarm.</i>
```
samthor, line 33: You could instead say here "Non-zero values create a repeating alarm that repeats every period."

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
+let display = document.querySelector('.alarm-display');
```
samthor, line 1: I know this is stylistic but I really prefer these to be `const`. They'll never change and it makes it much clearer looking at the code. I don't need you to go to all PRs and change this, just some advice.

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
...
+    alarmInfo.periodInMinutes = period;
+  }
+
+  // Create the alarm – this uses the same signature as chrome.alarms.create
+  manager.createAlarm(name, alarmInfo);
+});
+
+class AlarmManager {
+  constructor({display, log} = {}) {
+    if (display) {
```
samthor, line 48: these ifs don't add much. You always set these values in the demo. I'd also remove the `{}` args and just... have two args.

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
...
+      this.displayElement = display;
+    }
+    if (log) {
+      this.logElement = log;
+    }
+
+    this.logMessage('Manager: initializing demo');
+
+    this.displayElement.addEventListener('click', (event) => this.handleCancelAlarm(event));
```
samthor, line 57: Another way to do this is to define `handleCancelAlarm` as a property of this class, e.g.:

```js
  // instead of
  async handleCancelAlarm(event) { ... }

  // do this
  handleCancelAlarm = (event) => {};
```

That way it's already bound to `this` and your `addEventListener` calls are simpler.

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
...
+    let ms = pad(date.getMilliseconds(), 3);
+    let time = `${h}:${m}:${s}.${ms}`;
+
+    let logLine = document.createElement('div');
+    logLine.innerText = `[${time}] ${message}`;
+
+    // Log events in reverse chronological order
+    this.logElement.insertBefore(logLine, this.logElement.firstChild);
+
+    // Log events in chronological order
```
samthor, line 76: I'd remove this commented out bit.

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
...
+    // this.logElement.appendChild(logLine);
+  }
+
+  async handleAlarm(alarm) {
+    let json = JSON.stringify(alarm, null, 2).replace(/\s+/g, ' ');
```
samthor, line 81: I find this string replacement stuff a bit weird. I get it, you want your JSON to look nice—but it means an alarm called "hello     spaces" gets its spaces trimmed. I'd probably just use the already-quite-tight `JSON.stringify(alarm)` output. I don't think missing spaces really puts people off.

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
...
+    await this.cancelAlarm(name);
+    await this.refreshDisplay();
+  }
+
+  async cancelAlarm(name) {
+    // TODO: Remove custom promise wrapper once the Alarms API supports promises
+    return new Promise((resolve, reject) => {
+      chrome.alarms.clear(name, (wasCleared) => {
+        if (wasCleared) {
+          this.logMessage(`Manager: canceled alarm "${name}"`)
```
samthor, line 101: nit: missing trailing semicolon here and below

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
...
+  }
+
+  async cancelAllAlarms() {
+    // TODO: Remove custom promise wrapper once the Alarms API supports promises
+    return new Promise((resolve, reject) => {
+      chrome.alarms.clearAll((wasCleared) => {
+        if (wasCleared) {
+          resolve(wasCleared);
+        } else {
+          reject(wasCleared);
```
samthor, line 140: I get that for a demo an error here is useful, but I don't think the Promisified version would treat this as an error—it would simply resolve with false. If you felt like fixing that here and then throwing an error above—e.g.,

```js
cancelAllAlarms().then((result) => {
  if (!result) {
    throw new Error(`could not cancel alarms!`);
  }
})
```

—that might be nicer, but not required.

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
...
+        resolve();
+      });
+    });
+  }
+
+  // Simple locking mechanism to prevent multiple concurrent refreshes
+  #refreshing = false;
+
+  async refreshDisplay() {
+    if (this.#refreshing) return; // refresh in progress, bail
```
samthor, line 163: {}'s please

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
...
+
+    this.#refreshing = true;      // acquire lock
```
samthor, line 165: this should be

```js
this.#refreshing = true;
try {
  // do await here
} finally {
  this.#refreshing = false;
}
```

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
...
+    await Promise.all([
+      this.clearDisplay(),
+      this.populateDisplay(),
+    ]);
+    this.#refreshing = false;     // release lock
+
+    // Return undefined to avoid leaking implementation details
```
samthor, line 172: This comment and solo return statement are needless.

api/alarms/index.js
```diff
@@ -0,0 +1,182 @@
...
+    return;
+  }
+
+  async clearDisplay() {
+    this.displayElement.innerText = '';
```
samthor, line 177: innerText considered harmful. Always use `textContent`.

samthor (commented):
This is great. My main thoughts are that:

- it needs a short paragraph at the top of the file explaining what this demo does
- there needs to be a way to open the page again (just in case the user wants to play with the demo again)

My only other broader concern, and I don't want to bikeshed, is is it worth handling the alarms in the SW and announcing them to all pages? I think that's a more compelling reason to use alarms, not just for a single foreground page.

Alternatively, what happens if no listeners are registered while an alarm fires (i.e., this page is closed)? Would they be delivered to the next opening page? Is it worth listing that in the top note—"try adding an alarm, then closing the page, and coming back a minute later!".

api/alarms/background.js
dotproto: I skipped over that use case. It would be easiest to open a new tab on action click, so I'll run with that for the moment. 

This touches on a broader question, though: should we establish a set of conventions for extensions demos? I've opened #590 to kick around some ideas on this front.

api/alarms/index.html
dotproto: That feels a little long. How about "Repetition period"?

api/alarms/index.html
dotproto: Done!

api/alarms/index.js
dotproto: I'm amenable to that. This is just a stylistic thing I haven't adopted. Please ping me in the future if you see lapses.

api/alarms/index.js
dotproto: Ack.

api/alarms/index.js
dotproto: Ack.

api/alarms/index.js
dotproto: Done.

I originally added it 'cause I wanted to give developers the flexibility to adjust the output based on personal preference. If we do want to provide user customization for a published version of this, it would probably be best to expose that in the options page.

api/alarms/index.js
dotproto: Ooh, hadn't thought of that. I also updated the log CSS to prevent whitespace from being collapsed.

api/alarms/index.js
dotproto: Ack

api/alarms/index.js
dotproto: Thanks. I find try/catch in async functions to be awkward so my brain likes to gloss over it in some cases 😅

api/alarms/index.js
dotproto: TIL, will do

api/alarms/index.js
dotproto: Removed.

api/alarms/index.js
dotproto: Sorry, I don't follow this snippet. Are you suggesting that we throw in the event handlers declared on lines 11 and 12 or is this supposed to chain off of `chrome.alarms.clearAll`?

api/alarms/index.js
dotproto: Fixed.

api/alarms/index.html
samthor: Yes, that's better.

api/alarms/index.js
samthor: I'd change line 18 to this:

```js
clearButton.addEventListener('click', async () => {
  const wasCleared = manager.cancelAllAlarms();
  if (!wasCleared) {
    throw new Error('error clearing alarms');
  }
});
```

... but I would also be happy leaving it as-is now. My main point was like, the Promisified-version of the API will eventually resolve with a true or false boolean, which is what you've done now. So that's great.

samthor (approved):

api/alarms/index.js
dotproto: For the sake of reducing churn I'm going to stick with what I have now.

PR #589, opened 2021-04-16
Action api nits

dotproto:
I forgot to commit two minor changes before merging #574


dotproto:
Using admin override to merge this as it's only whitespace cleanup.

PR #595, opened 2021-05-14
Fix Uncaught TypeError: $(...).empty(...).dialog is not a function

figureConan:
Uncaught TypeError: $(...).empty(...).append(...).dialog is not a function


alliancenshuti (changes requested):
Hi there, to what am able to see I am wondering why do we have to import jquery on line 3 and 4 and if one of them isn't helping why don't we remove it to make things cleaner and clear. thanks

samthor (approved):

PR #612, opened 2021-07-12
Text Replacer example

dotproto:
Adds a sample extension called "Text Replacer." As the name implies, this extension will replace a string in the page with a custom string. Currently supports 5 replacements.


examples/text-replacer/content.js
```diff
@@ -0,0 +1,47 @@
...
+    if (!source[i]) { continue; }
+    const [find, replace] = source[i];
+    const sanitizedMatch = escapeRegExp(find);
+    const findExp = new RegExp(`\\b${sanitizedMatch}\\b`, 'gi');
+    output[i] = [findExp, replace];
+  }
+  return output;
+}
+
+// Use var to avoid "Identifier 'REGEXP_SPECIAL_CHARACTERS' has already been
```
samthor, line 25: Oh wow. I didn't realise this runs non-ESM code and runs in the top-scope.

Everything in this file is being redeclared always. Would it be worth putting it in an IIFE? Is that too complex?

examples/text-replacer/content.js
```diff
@@ -0,0 +1,47 @@
...
+// declared" errors when running multiple times on the same page.
+var REGEXP_SPECIAL_CHARACTERS = /[.(){}^$*+?[\]\\]/g;
+/** Sanitize user input to prevent unexpected behavior during RegExp execution */
+function escapeRegExp(pattern) {
+  return pattern.replace(REGEXP_SPECIAL_CHARACTERS, "\\$&")
```
samthor, line 30: I highly suspect this might not be all regex chars. Maybe call out that this is for example purposes only?

examples/text-replacer/popup.html
```diff
@@ -0,0 +1,80 @@
...
+      </tbody>
+    </table>
+
+    <div class="form--controls">
+      <input id="clear" type="button" value="Clear">
+      <input id="submit" type="submit" value="Replace">
+    </div>
+  </form>
+
+  <script src="popup.js"></script>
```
samthor, line 77: Can you use `type="module"`? Then you can use top-level await in the code, rather than the awkward loading and call to `loadFormData`.

examples/text-replacer/popup.js
```diff
@@ -0,0 +1,130 @@
...
+  await saveFormData();
+
+  let currentTab = await getCurrentTab();
+  chrome.scripting.executeScript({
+    target: {tabId: currentTab.id},
+    files: ['content.js']
+  });
+});
+
+document.getElementById('clear').addEventListener('click', (event) => {
```
samthor, line 24: You can get rid of this and change the `<button>` to be an `<input type="reset" />`.

examples/text-replacer/popup.js
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/**
+ * Write the form values to persistent storage.
+ */
+function saveFormData() {
+  const inputs = form.querySelectorAll('input[type=text]');
+
+  const patterns = [...inputs].reduce((acc, input, index) => {
+    const outerIndex = index >> 1;
```
samthor, line 50: Do a real division and round it. Sorry, but we should be clear to novices.

examples/text-replacer/content.js
```diff
@@ -0,0 +1,47 @@
+// Copyright 2021 Google LLC
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
+
+// Replace text on the page using a static list of patterns
+function textReplacer(replacements) {
```
samthor, line 8: I think having this method is a bit confusing. I'd just move its two lines into the callback for the fetcher, and expand the comment:

```js
// Get the patterns to replace from storage, then build a regex from them and replace all text on the page.
```

samthor (approved):

examples/text-replacer/background.js
```diff
@@ -0,0 +1,29 @@
+chrome.commands.onCommand.addListener((command, tab) => {
+  if (command == 'replace-text') {
+    replaceText(tab.id);
+  }
+});
+
+chrome.runtime.onInstalled.addListener(() => {
+  registerContextMenus();
+});
```
rdcronin, line 9: During extension update, will this cause an error by trying to re-register context menu items?

examples/text-replacer/background.js
```diff
@@ -0,0 +1,29 @@
...
+
+function registerContextMenus() {
+  chrome.contextMenus.create({
+    id: 'replace-text-menuitem',
+    title: 'Replace text',
+  });
+}
+
+chrome.contextMenus.onClicked.addListener((info, tab) => {
+  if (info.menuItemId == 'replace-text-menuitem') {
```
rdcronin, line 19: extract "replace-text-menuitem" into a constant

examples/text-replacer/background.js
```diff
@@ -0,0 +1,29 @@
...
+    replaceText(tab.id);
+  }
```
rdcronin, line 21: I'd normally recommend throwing an assert in here, since this is the only id we ever expect.  Similarly for commands.

examples/text-replacer/content.js
```diff
@@ -0,0 +1,47 @@
...
+  const replacementPatterns = buildReplacementRegex(replacements);
+  replaceText(replacementPatterns);
+}
+
+function buildReplacementRegex(source) {
+  const output = [];
+  for (var i = 0; i < source.length; i++) {
```
rdcronin, line 15:
```suggestion
  for (let i = 0; i < source.length; i++) {
```

examples/text-replacer/content.js
rdcronin: +1

examples/text-replacer/content.js
```diff
@@ -0,0 +1,47 @@
...
+/** Sanitize user input to prevent unexpected behavior during RegExp execution */
+function escapeRegExp(pattern) {
+  return pattern.replace(REGEXP_SPECIAL_CHARACTERS, "\\$&")
+}
+
+/** Iterate through all text nodes and replace  */
+function replaceText(replacements) {
+  let node;
+  const nodeIterator = document.createNodeIterator(document.body, NodeFilter.SHOW_TEXT);
+  while (node = nodeIterator.nextNode()) {
```
rdcronin, line 37: can we declare node here?  We only need it in the body

examples/text-replacer/content.js
```diff
@@ -0,0 +1,47 @@
...
+    for (let [find, replace] of replacements) {
+      node.nodeValue = node.nodeValue.replace(find, replace);
+    }
+  }
+}
+
+// Replace text on the page using a list of patterns loaded from storage
+chrome.storage.sync.get(['patterns'], function(data) {
+  textReplacer(data.patterns);
+});
```
rdcronin, line 47: We're going to be gradually encouraging developers to avoid storage access by content scripts unless crucial.  WDYT about changing this to a message to the background page?  (I don't feel strongly)

examples/text-replacer/manifest.json
```diff
@@ -0,0 +1,35 @@
...
+    "storage",
+    "commands",
+    "contextMenus"
+  ],
+  "background": {
+    "service_worker": "background.js"
+  },
+  "commands": {
+    "replace-text": {
+      "description": "Replace text on the current page."
```
rdcronin, line 32: Doesn't this need a default keybinding?

examples/text-replacer/manifest.json
```diff
@@ -0,0 +1,35 @@
...
+  "version": "1.0.0",
+  "manifest_version": 3,
+  "description": "Replace any a line of text on a page with another line of text.",
+  "icons": {
+    "32": "icons/icon32.png",
+    "48": "icons/icon48.png",
+    "64": "icons/icon64.png"
+  },
+  "action": {
+    "default_title": "Show an alert",
```
rdcronin, line 12: Is this an alert?

nagayev:
+1
It was useful for me.

patrickkettner:
@sebastianbenz @oliverdunk what should we do with these older PRs?

patrickkettner:
@dotproto, since we still dont have a textreplacer sample, any chance youd like to update this to get it merged?

examples/text-replacer/background.js
dotproto: Yes, it would. I think I wrote it this way originally because I considered updates to be out of scope. I've updated the code as suggested in order to better set the reader up for success.

Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/background.js
dotproto: Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/background.js
dotproto: Added explicit errors for unknown commands and context menus.

Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/content.js
dotproto: Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/content.js
dotproto: Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/content.js
dotproto: I've moved the entire script into an IIFE and added a comment explaining why. 

Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/content.js
dotproto: Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/content.js
dotproto: No. Unlike `for-in` statements, which may contain declarations, `while` statements require an expression in the condition position.  We also can't use `for-in` here because the `NodeIterator` interface predates JavaScript's [iteration protocol](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Iteration_protocols).  The closest we can do is move the `let` declaration next to the `while` to make their relationship more clear. I've applied this change to my new PR. 

Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/content.js
dotproto: Updated the example to route all read/writes through the background context.

Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/manifest.json
dotproto: Oops.  Sloppy copy-paste from another example I was referencing.


Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/manifest.json
dotproto: Strictly speaking, no, a default keybinding is not required. The user could still manually assign a key binding if they wanted to use it. 

For the sake of clarity, I've updated the demo to have a default keyboard shortcut.

Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/popup.js
dotproto: I've replaced the button with `<input type="reset">`, but still need an event listener in order to persist the changes. 

Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

examples/text-replacer/popup.js
dotproto: Addressed in https://github.com/GoogleChrome/chrome-extensions-samples/pull/1718.

dotproto:
I'm no longer able to edit the `text-replacer` brach, so this PR has been replaced by #1718.

PR #639, opened 2021-10-08
Fix generating setBadgeBackgroundColor() parameter

pkotwicz:
This fixes the "Randomize badge background color" <button>


oliverdunk:
It looks like this was fixed already, so I'm going to close this out. Sorry for the slow reply and thanks regardless for the contribution.

PR #649, opened 2021-11-14
correction: hello world example contains call to onInstalled, not onC…

mattheson:


mattheson:
sorry, forgot to submit CLA. will make new pr

PR #650, opened 2021-11-14
correction: hello world example contains call to onInstalled, not onClicked

mattheson:


dotproto (approved):
Thanks for catching this!

PR #651, opened 2021-11-24
Fix page redder to disregard any chrome internal pages

codetheorist:
The page redder example currently affects any tab, regardless of the URL, which can result in errors when this is attempted on an internal Chrome page.

This fix prevents the extension from affecting any Chrome internal page.

### DIFF
--- a/examples/page-redder/background.js
+++ b/examples/page-redder/background.js
@@ -3,8 +3,10 @@ function reddenPage() {
 }
 
 chrome.action.onClicked.addListener((tab) => {
-  chrome.scripting.executeScript({
-    target: { tabId: tab.id },
-    function: reddenPage
-  });
+  if(!tab.url.includes("chrome://")) {
+    chrome.scripting.executeScript({
+      target: { tabId: tab.id },
+      function: reddenPage
+    });
+  }
 });


examples/page-redder/background.js
```diff
@@ -3,8 +3,10 @@ function reddenPage() {
 }
 
 chrome.action.onClicked.addListener((tab) => {
-  chrome.scripting.executeScript({
-    target: { tabId: tab.id },
-    function: reddenPage
-  });
+  if(!tab.url.includes("chrome://")) {
```
dotproto, line 6: I _think_ this should only match against the URL's scheme, but I think it would be a bit safer to replace `includes()` with `startsWith()` in order to avoid situations where this string appears in other parts of the URL. 

```suggestion
  if(!tab.url.startsWith("chrome://")) {
```

dotproto (approved):
Thanks for this. I had one minor suggestion I'd like to see, but given how long I've taken to review this I don't think it's worth blocking the PR.

alexherbo2:
We can also do `new URL(tab.url).protocol`.

Reference: https://developer.mozilla.org/en-US/docs/Web/API/URL

PR #664, opened 2022-01-22
fix: fix Uncaught TypeError in action api demo

boymax1:


dotproto (approved):
Oops, thanks for catching that!

PR #678, opened 2022-03-17
Convert water-alarm-notification to MV3

AmySteam:
This PR proposes the following:

- MV3 example for the Drink Water extension example.
- Placed extension under **examples** instead of **API** since there's already a better alarms API demonstration. Thoughts?
- Updated code to ES6
- Changed sample time from 1 second to 1 minute (Alarms API has this limitation)
- Removed a link in popup.html pointing to CSP docs for [Google Apps](https://developer.chrome.com/docs/apps/contentSecurityPolicy/) 

@dotproto This example will be used in [Design the UI](https://developer.chrome.com/docs/extensions/mv3/user_interface/)


dotproto (approved):
I only have a couple nits that I'll clean up in a separate PR. Thanks for updating this, @AmySteam!

PR #679, opened 2022-03-17
Convert Tab Flipper to MV3

AmySteam:
- Converts Tab Flipper commands API example to MV3.
- Updated code to ES6

@dotproto This example will be used in [Design the UI](https://developer.chrome.com/docs/extensions/mv3/user_interface/)


dotproto (approved):
I only have a couple nits that I'll clean up in a separate PR. Thanks for updating this, @AmySteam!

PR #680, opened 2022-03-18
Convert Global Google Search to MV3

AmySteam:
- Converted Global Google Search to MV3
- Refactored code to ES6
- Options page and popup were the same. I left the popup and removed the options page.

@dotproto This example will be used in [Design the UI ](https://developer.chrome.com/docs/extensions/mv3/user_interface/)


dotproto (approved):
I had a few nits in an old review I lost track of, but none of them are blocking and I'm already planning to do some cleanup. As such, I think we're good to merge as is.

PR #705, opened 2022-04-19
Getting Started Tutorial: Tabs Manager

AmySteam:
[Tutorial Scope](https://docs.google.com/document/d/1e6qhY9s3dllNH1Xwu41fZh18hIe8ZwewrWjGkxi-CCQ/edit?usp=sharing)

This extension example uses the tabs API in a popup. It lists all the extension docs that are open and can group these tabs.


dotproto (changes requested):
Just nits. Once these are addressed I think we're good to merge.

AmySteam:
@dotproto I applied your suggestions. It's ready to merge!

dotproto (approved):
Looks good. Thanks, @AmySteam!

tutorials/tabs-manager/popup.html
```diff
@@ -0,0 +1,25 @@
...
+          <p class="pathname">Tab Pathname</p>
+        </a>
+      </li>
+    </template>
+
+    <h1>Google Dev Docs</h1>
+    <button>Group Tabs</button>
+    <ul></ul>
+
+    <script src="./popup.js" type="module"></script>
```
alexherbo2, line 23: @AmySteam Although it is not specific to Chrome extensions, I discovered we can use [top level await] in the background page. It could be worth to highlight it in the tutorial. What do you think?

[Top level await]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#top_level_await

AmySteam: @alexherbo2 That's a great idea! I'll include it in the tutorial 👍 

I really appreciate it that you are taking time to share your ideas. Please feel free to comment on any of the tutorials :)

PR #720, opened 2022-06-09
Add tabs permission

AOSXAP:
A new tab cannot be opened unless the tabs permission is enabled.

### DIFF
--- a/examples/hello-world/manifest.json
+++ b/examples/hello-world/manifest.json
@@ -1,9 +1,10 @@
 {
-  "name": "Hello, World!",
-  "version": "1.0",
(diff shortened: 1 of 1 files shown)


dotproto:
Closing this PR as the tabs permission is not required to call `chrome.tabs.create()`.

As a more general note, this PR changes the formatting for the affecte file. Generally speaking it's a best practice to limit changes in a PR to only what is necessary to implement the change you're trying to make. In this case, that would be only inserting `"permissions" : ["tabs"],` on line 5. Also, when doing so I'd strongly recomend matching the formatting of the existing code.

PR #729, opened 2022-07-01
Fix: Switch tab and window update

AmySteam:
Fixes bug that focused on the Window before the tab.

I switched the order and works as expected.


dotproto (approved):
Thanks, @AmySteam!

boontham33345:
Good

PR #746, opened 2022-08-20
Created MV3 Iframe sandBox Example

Ambushfall:


Ambushfall:
Port of MV2's iFrame / Notification to MV3

Also added a showcase of a fetch for random JSON endpoint and it's data visible in the console.

Ambushfall:
Fixed API/action demo failing to reset the icon.

dotproto (changes requested):
Thanks for your interest in creating a sample extension. There are a few issues with this PR that will need to be addressed before we can accept your contribution.

At the moment the biggest issue is that the "sandbox" sample introduced by this PR doesn't fit into the kinds of samples we're collecting in this repo. We currently have two main types of samples:

* The [`api`](https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/api) directory contains extensions that show as much of the API surface of a given namespace as possible in a single demo.
* The [`example`](https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/api) directory contains fully functional extensions that use a variety of APIs to achieve their goals. That said, they may not be as polished as something you'd find on the Chrome Web Store.

The sample in this PR doesn't use enough of the API surface for the `api` directory and doesn't do enough to fit in the  `example` directory. I'm also currently considering introducing a third type of sample:

* The `cookbook` directory collects functional samples that demonstrate a specific concept or way of accomplishing a given task.

The `cookbook` approach seems to be the best for your sandbox sample. Before discussing the changes necessary to make this a good cookbook sample, though, there are two things I'd like to get cleaned up:

1. Please revert all changes to files in `api/action`. These changes seem unintentional and are out of scope for introducing a new sample.
2. The sandbox sample contains code that fetches data from a remote server, but this data doesn't meaningfully contribute to the example being presented. Please update the sample to either integrate this data or remove the unused code.

PR #756, opened 2022-09-18
Fix code samples for popup script due to Manifest v3 update

pwwpche:
According to https://developer.chrome.com/docs/extensions/reference/scripting/#type-ScriptInjection, the property name for function injection has been changed to `func` from `function`. This change update the sample script accordingly.


dotproto (approved):
LGTM, thanks!

PR #771, opened 2022-11-02
Fix favicon API example

### LINKED ISSUE #770: Favicon API is wrong
**Describe the bug**

Favicon example here is incorrect: https://github.com/GoogleChrome/chrome-extensions-samples/blob/2938c523066aa0b5903945999192afbe76167277/api/favicon/content.js

Going through the issue tracker, it looks like this API was changed and the changes are not reflected in the example. Also this is one of the only sources of documentation for this API, so it's important to get it right.

Also, it would be good to have some notes somewhere about what happens when the favicon isn't currently cached. Does it reach out to the site and get it? What if the page can't be retrieved -- does it fall back to the domain's favicon?

**To Reproduce**

Using the example results in: 

image

There are two problems:
* page_url should be pageUrl
* the pageUrl parameter should be escaped. Using the URL object's searchParms makes this easy.

**Expected behavior**

This is a working example:

```
function faviconURL(u) {
  const url = new URL(chrome.runtime.getURL("/_favicon/"));
  url.searchParams.set("pageUrl", u); // this encodes the URL
  url.searchParams.set("size", "32");
  return url.toString();
}
```

**Notes**

PR will follow

hut8:
Resolves #770

### DIFF
--- a/api/favicon/content.js
+++ b/api/favicon/content.js
@@ -1,6 +1,12 @@
+function faviconURL(u) {
+  const url = new URL(chrome.runtime.getURL("/_favicon/"));
+  url.searchParams.set("pageUrl", u); // this encodes the URL as well
+  url.searchParams.set("size", "64");
+  return url.toString();
+}
+
 window.onload = e => {
   const img = document.createElement('img');
-  const url = '_favicon/?page_url=https://www.google.com&size=64';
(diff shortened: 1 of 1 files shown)


dotproto (approved):
Thanks for fixing this!

Issue #795, opened 2022-12-25
Impossible to write to or read from clipboard in offscreen document


guest271314:
**Describe the bug**
Impossible to read from or write to clipboard in an offscreen document - even when `clipboard_read` and `clipboard_write` permissions are `"granted"`.

**To Reproduce**

Navigate to Details of the extension, then Site settings, and set Clipboard permission to Allow.

```
await chrome.offscreen.createDocument({
  url: 'offscreen.html',
  justification: 'ignored',
  reasons: ['CLIPBOARD'],
});
```

In the script loaded by the offscreen document 

```
(async () => {
  try {
    let clipboard_permissions = (
      await navigator.permissions.requestAll([
        { name: 'clipboard-read' },
        { name: 'clipboard-write' },
      ])
    ).map(({ state }) => state);
    if (clipboard_permissions.every((state) => state === 'granted')) {
      await navigator.clipboard.writeText(location.href);
      console.log(await navigator.clipboard.readText(location.href));
    } else {
      console.log(clipboard_permissions);
    }
  } catch (e) {
    console.error(e.message);
  }
})();
```

```
Document is not focused.
```

**Expected behavior**
Read and write to clipboard in an offscreen document.

**Screenshots**

**Notes**
I don't think it is possible to gain focus on an offscreen document. At least not using any of the approaches I have tried, including `focus()` and calling the clipboard read and write methods in an `NotificationEvent` handler.

Nonetheless `CLIPBOARD` is listed as a "reason" in offscreen documentation https://developer.chrome.com/docs/extensions/reference/offscreen/, though appears to be useless at this point.

dotproto:
As you indicated, the web Clipboard API can only be used in the currently focused page. This is limitation of the Clipboard API was intentional for the web platform, but may be something we can change for extensions platform. I just opened [crbug.com/1404835](https://crbug.com/1404835) to track this feature request. That issue also contains link sot several other related issues, the most notable of which is probably [1160302](https://crbug.com/1160302), which is a request to expose the clipboard API to extension service workers.

More germain to your specific request, it _is_ possible to interact with the system clipboard using `document.execComand('copy')`. PR #797 (opened earlier today) is a cookbook example showing how to use it in an offscreen document.

guest271314:
[...]

Issue #781, opened 2022-11-26
Sample: Assign arbitrary Window, Worker, SharedWorker, Worklet as a Client


guest271314:
Sample: Assign an arbitrary `Window`, `Worker`, `SharedWorker`, `Worklet` as a `Client` of MV3 `ServiceWorker`.

This will eliminate the current need for multiple workarounds to establish communication between the `ServiceWorker` and arbitrary Web page for the ability to _transfer_ Transferable Objects, etc. from the `ServiceWorker` to the arbitrary Web page.

dotproto:
Closing as this appears to be an extension platform feature request rather than an extension sample one.

To more directly address the request, this feels like something Chrome is very unlikely to do for two primary reasons. First, this type of use is well outside the intended design and utility of service workers and their clients. Second, this use case is would be more effectively addressed another way, like moving the extension platform from a custom message system to the web platform messaging model.

Please open an issue on crbug.com. Ideally, this issue report should focus on the problem you're encountering rather than a specific solution.

Issue #778, opened 2022-11-22
Sample: Set opener on Window created with chrome.windows.create()


guest271314:
`setSelfAsOpener` option of `chrome.windows.create()` https://developer.chrome.com/docs/extensions/reference/windows/#method-create throws from MV3 `ServiceWorker`

```
Uncaught (in promise) Error: Cannot specify setSelfAsOpener Service Worker extension.
```
and is not defined in content scripts.

Sample request:  Set `opener` on `Window` created with `chrome.windows.create()`.

guest271314:
There is currently no way to set `opener` for the `Window` created by `chrome.offscreen.createDocument()` either.

Sample request: Programmatically set `opener` on `Window` created with `chrome.offscreen.createDocument()`.

patrickkettner:
Closing this out, feel free to follow along at https://crbug.com/1275145

Issue #766, opened 2022-10-27
Sample: Transfer a blob from a background context to a page


dotproto:
**Goal of the demo**

Demonstrate how an extension can create a blob in the extension's service worker and transfer it to a page's main world. 

**Suggested implementation**

Sketch of my initial implementation plan

1. Have a content script inject an iframe into a page
2. Run a script in the iframe to post a message back to the service worker using `navigator.serviceWorker.controller.postMessage()`
3. In the SW's message handler, generate a blob OR use `structuredClone()` to create a copy of a blob.
4. Use `event.source.postMessage(msg, [transferable])` to reply to the client and transfer the blob
5. In the iframe's message handler, use `window.parent.postMessage(msg, "<origin>", [transferable])` to transfer the data to the page

**Related links**

* https://github.com/w3c/webextensions/issues/293

**Notes**

Initial findings suggest that it's not possible to transfer variables across origins. I'm tentatively thinking that to work around this, we can use `URL.createObjectURL()` in the iframe and directly reference the object URL somewhere that's easily visible to the end user (e.g. canvas, Audio/Video element, etc.).

guest271314:
Here you go https://github.com/guest271314/persistent-serviceworker/tree/main/chromium_extension_web_accessible_resources_iframe_message_event. You can transfer whatever you want. Bonus: Keeps MV3 `ServiceWorker` persistent.

guest271314:
Note, the last time I checked `Blob`s are not transferable.

guest271314:
Details of the message passing. We create a `MessageChannel` pair in the MV3 `ServiceWorker`, then transfer one of the `MessagePort`s created to the arbitrary Web page where action icon is clicked, to establish direct communication between `SeriveWorker` and the Web page where `port` is defined globally. 

To send a `Blob` from the `ServiceWorker` to the Web page where action icon is clicked include on Line 111 of background.js

```
port1.postMessage(new Blob(['Message from MV3 ServiceWorker'], {type:'text/plain'})); 
```
To log messages in the `ServiceWorker` sent from `port2` defined globally in the Web page where click action occurred add the following to Line 114 in background.js

```
  console.log(e.data);
```

To transfer substantial amounts of data we can transfer `ArrayBuffer` (and `TypedArray` `buffer`), `WebAssembly.Memory`, and on Chromium and Chrome Transferable Streams, e.g.,

```
const {readable,  writable} = new TransformStream();
[...]

PR #797, opened 2023-01-03
Add offscreen clipboard cookbook example

dotproto:
This PR contains a functional demo of how to interact with the system clipboard using the [Offscreen API](https://developer.chrome.com/docs/extensions/reference/offscreen) and web APIs (namely [`document.execCommand()`](https://developer.mozilla.org/en-US/docs/Web/API/Document/execCommand), [MDN guide](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard)).

### DIFF
--- a/cookbook/offscreen-clipboard-write/background.js
+++ b/cookbook/offscreen-clipboard-write/background.js
@@ -15,20 +15,21 @@
 // The value that will be written to the clipboard.
 const textToCopy = `Hello world!`;
 
-// When the browser action is clicked, `addToClipboard()` will use an offscreen document to write
-// the value of `textToCopy` to the system clipboard.
+// When the browser action is clicked, `addToClipboard()` will use an offscreen
+// document to write the value of `textToCopy` to the system clipboard.
 chrome.action.onClicked.addListener(async () => {
   await addToClipboard(textToCopy);
 });
 
-// Solution 1 - As of Jan 2023, service workers cannot directly interact with the system clipboard.
-// To work around this, we'll create an offscreen document and pass it the data we want to write to
-// the clipboard.
+// Solution 1 - As of Jan 2023, service workers cannot directly interact with
+// the system clipboard. To work around this, we'll create an offscreen document
+// and pass it the data we want to write to the clipboard.
 async function addToClipboard(value) {
-  // This pattern show in this if-else ensures that the offscreen document exists before we try to
-  // send it a message. If we didn't await the `hasDocument()` and `createDocument()` calls, the
-  // `sendMessage()` call could be sent before the offscreen document can register it's
-  // `runtime.onMessage` listener.
+  // This pattern ensures that the offscreen document exists before we try to
+  // send it a message. If we didn't await the `hasDocument()` and
+  // `createDocument()` calls, the `sendMessage()` call could send a message
+  // before the offscreen document can register it's `runtime.onMessage`
+  // listener.
   if (await chrome.offscreen.hasDocument()) {
     console.debug('Offscreen doc already exists.');
   } else {
@@ -41,7 +42,8 @@ async function addToClipboard(value) {
     });
   }
 
-  // Now that are sure we have an offscreen document, we can safely dispatch the message.
+  // Now that are sure we have an offscreen document, we can safely dispatch the
+  // message.
   chrome.runtime.sendMessage({
     type: 'copy-data-to-clipboard',
     target: 'offscreen-doc',
--- a/cookbook/offscreen-clipboard-write/background.js
+++ b/cookbook/offscreen-clipboard-write/background.js
@@ -12,28 +12,28 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// TODO(dotproto): Add more comments explaining this code.
-
+// The value that will be written to the clipboard.
 const textToCopy = `Hello world!`;
 
+// When the browser action is clicked, `addToClipboard()` will use an offscreen document to write
+// the value of `textToCopy` to the system clipboard.
 chrome.action.onClicked.addListener(async () => {
   await addToClipboard(textToCopy);
 });
 
-// Solution 1 - As of Jan 2023, service workers cannot directly interact with
-// the system clipboard. To work around this, we'll create an offscreen document
-// and pass it the data we want to write to the clipboard.
-
+// Solution 1 - As of Jan 2023, service workers cannot directly interact with the system clipboard.
+// To work around this, we'll create an offscreen document and pass it the data we want to write to
+// the clipboard.
 async function addToClipboard(value) {
+  // This pattern show in this if-else ensures that the offscreen document exists before we try to
+  // send it a message. If we didn't await the `hasDocument()` and `createDocument()` calls, the
+  // `sendMessage()` call could be sent before the offscreen document can register it's
+  // `runtime.onMessage` listener.
   if (await chrome.offscreen.hasDocument()) {
-    console.debug('Offscreen doc already exists');
+    console.debug('Offscreen doc already exists.');
   } else {
     console.debug('Creating a new offscreen document.');
 
-    // TODO(dotproto): Follow up with eng to make sure that the doc's
-    // `runtime.onMessage` listener will be registered before the
-    // `createDocument()` call resolves. If not, we'll need to rewrite this to
-    // guarantee the correct order of operations.
     await chrome.offscreen.createDocument({
       url: 'offscreen.html',
       reasons: [chrome.offscreen.Reason.CLIPBOARD],
@@ -41,6 +41,7 @@ async function addToClipboard(value) {
     });
   }
 
+  // Now that are sure we have an off document, we can safely dispatch the message.
   chrome.runtime.sendMessage({
     type: 'copy-data-to-clipboard',
     target: 'offscreen-doc',
@@ -49,8 +50,7 @@ async function addToClipboard(value) {
 }
 
 // Solution 2 – Once extension service workers can use the Clipboard API,
-// replace the offscreen document based implementation with this.
-
+// replace the offscreen document based implementation with something like this.
 async function addToClipboardV2(value) {
-  navigator.clipboard.copyText(value)
+  navigator.clipboard.writeText(value)
 }
--- a/cookbook/offscreen-clipboard-write/background.js
+++ b/cookbook/offscreen-clipboard-write/background.js
@@ -52,5 +52,5 @@ async function addToClipboard(value) {
 // Solution 2 – Once extension service workers can use the Clipboard API,
 // replace the offscreen document based implementation with something like this.
 async function addToClipboardV2(value) {
-  navigator.clipboard.writeText(value)
+  navigator.clipboard.writeText(value);
 }
--- a/cookbook/offscreen-clipboard-write/background.js
+++ b/cookbook/offscreen-clipboard-write/background.js
@@ -41,7 +41,7 @@ async function addToClipboard(value) {
     });
   }
 
-  // Now that are sure we have an off document, we can safely dispatch the message.
+  // Now that are sure we have an offscreen document, we can safely dispatch the message.
   chrome.runtime.sendMessage({
     type: 'copy-data-to-clipboard',
     target: 'offscreen-doc',
--- a/cookbook/offscreen-clipboard-write/background.js
+++ b/cookbook/offscreen-clipboard-write/background.js
@@ -22,8 +22,9 @@ chrome.action.onClicked.addListener(async () => {
 });
 
 // Solution 1 - As of Jan 2023, service workers cannot directly interact with
-// the system clipboard. To work around this, we'll create an offscreen document
-// and pass it the data we want to write to the clipboard.
+// the system clipboard using either `navigator.clipboard` or
+// `document.execCommand()`. To work around this, we'll create an offscreen
+// document and pass it the data we want to write to the clipboard.
 async function addToClipboard(value) {
   // This pattern ensures that the offscreen document exists before we try to
   // send it a message. If we didn't await the `hasDocument()` and
--- a/cookbook/offscreen-clipboard-write/background.js
+++ b/cookbook/offscreen-clipboard-write/background.js
@@ -1,0 +1,53 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+// TODO(dotproto): Add more comments explaining this code.
+
+const textToCopy = `Hello world!`;
+
+chrome.action.onClicked.addListener(async () => {
+  await addToClipboard(textToCopy);
+});
+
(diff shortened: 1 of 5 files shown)


cookbook/offscreen-clipboard-write/offscreen.html
```diff
@@ -0,0 +1,3 @@
+<!DOCTYPE html>
+<input id="text" type="text">
```
tophf, line 2: A `textarea` is always used for this task because it preserves linebreaks inside the text.

dotproto: Good call, thanks!

```suggestion
<textarea id="text" type="text"></textarea>
```

IanStanion-google:
@dotproto I added some extra instructional comments to the code, as well as made the call to leave the additional  verification steps because I think they add a lot of instructional value.

IanStanion-google (approved):

cookbook/offscreen-clipboard-write/offscreen.html
```diff
@@ -0,0 +1,3 @@
+<!DOCTYPE html>
+<textarea id="text" type="text"></textarea>
```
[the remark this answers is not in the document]
dotproto: I'll name this on trying to snipe in a fix through the GH UI late at night. Thanks, @tophf :)

```suggestion
<textarea id="text"></textarea>
```

cookbook/offscreen-clipboard-write/offscreen.js
```diff
@@ -12,7 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-// TODO(dotproto): Add more comments explaining this code.
+// Once the message has been posted from the service worker, checks are made to confirm the message type and target before proceeding. 
+// This is so that the module can easily be adapted into existing workflows where secondary uses for the document (or alternate offscreen documents) might be implemented.
```
dotproto, line 16:
```suggestion
// Once the message has been posted from the service worker, checks are made to confirm the message
// type and target before proceeding. This is so that the module can easily be adapted into existing
// workflows where secondary uses for the document (or alternate offscreen documents) might be
// implemented.
```

cookbook/offscreen-clipboard-write/offscreen.js
```diff
@@ -28,6 +29,8 @@ async function handleMessages(message) {
   }
 }
 
+// The handleClipboardWrite function makes use of a DOM functionality API (document), which is currently impossible from inside a service worker.
```
dotproto, line 32:
```suggestion
// The `handleClipboardWrite` function makes use of a DOM functionality API (document), which is
// currently impossible from inside a service worker.
```

cookbook/offscreen-clipboard-write/offscreen.js
```diff
@@ -0,0 +1,47 @@
...
+    handleClipboardWrite(message.data);
+  }
+}
+
+// The handleClipboardWrite function makes use of a DOM functionality API (document), which is currently impossible from inside a service worker. 
+
+async function handleClipboardWrite(data) {
+  // Return early if we received the wrong kind of data.
+  if (typeof data !== 'string') {
+    console.debug(`Unexpected data value. Expected a 'string' value but received '${typeof data}'. Aborting.`)
```
dotproto, line 37:
```suggestion
    console.debug(`Unexpected data value. Expected a value with type 'string' but received a value with type '${typeof data}'. Aborting.`)
```

cookbook/offscreen-clipboard-write/background.js
```diff
@@ -0,0 +1,53 @@
...
+    type: 'copy-data-to-clipboard',
+    target: 'offscreen-doc',
+    data: value,
+  });
+}
+
+// Solution 2 – Only use this when
+
+async function addToClipboardV2(value) {
+  navigator.clipboard.copyText(value)
```
guest271314, line 55: `copyText` is not a method oc `navigator.clipboard`. 

The idea is to utilize Async Clipboard API to write a File to the clipboard with ClipboardItem and get the contents of clipboard items on arbitrary Web pages that can read from the clipboard. Hence the use of Async Clipboard API in https://github.com/GoogleChrome/chrome-extensions-samples/issues/795. This example does not achieve that.

guest271314 (changes requested):

cookbook/offscreen-clipboard-write/offscreen.js
```diff
@@ -17,34 +17,46 @@
...
+
+  // `document.execCommand('copy')` works against the user's selected in a web page. As such, we
+  // need to insert the updated text into the document and select it before it can be added to the
+  // clipboard.
   textEl.value = data;
   textEl.select();
   document.execCommand('copy');
 }
 
+// This textarea is used to easily modify and select it's text value.
```
guest271314, line 61: The point of filing https://github.com/GoogleChrome/chrome-extensions-samples/issues/795 was to explore another way to solve https://github.com/GoogleChrome/chrome-extensions-samples/issues/766 by writing a `File` to the clipboard, and get the `File` on arbitrary Web pages - not to paste plain text to the clipboard.

cookbook/offscreen-clipboard-write/background.js
dotproto: Thanks for pointing out the mistake. Looks like I accidentally swapped "write" for "copy". This has been addressed in the latest commit.

The original issue reported in #795 it explicitly describes writing text to and reading text from the clipboard. I'm not against adding a file demo, but that's out of scope for what we're showing in this sample.

guest271314: The original issue used _async_ Clipboard API for a reason. Perhaps I should have been very explicit about what I was trying to do.

Should I file another issue explaining precisely that the requirement is to write a `File` to the clipboard from an offscreen document - which _is_ currently impossible?

dotproto: Yep, that would be helpful for tracking and scheduling purposes. Note that creation of an extension sample for this is currently blocked by [1404835](https://bugs.chromium.org/p/chromium/issues/detail?id=1404835) (or another appropriate crbug).

cookbook/offscreen-clipboard-write/offscreen.js
dotproto: Resolving this conversation as it's a duplicate of the discussion in [this comment thread](https://github.com/GoogleChrome/chrome-extensions-samples/pull/797#discussion_r1063973635).

cookbook/offscreen-clipboard-write/background.js
guest271314: I just tried this 

```
async function writeClipImg() {
  try {
    //const imgURL = "https://avatars.githubusercontent.com/u/1047763?s=40&v=4";
    //const data = await fetch(imgURL);
    const blob = new Blob(['data'],{
      type: 'application/octet-stream'
    });

    await navigator.clipboard.write([new ClipboardItem({
      [blob.type]: blob,
    }), ]);
    console.log("Fetched image copied.");
  } catch (err) {
    console.error(err.name, err.message);
  }
}
onclick = async(e)=>{
  onclick = null;
  try {
    await writeClipImg();
    // async function getClipboardContents() {

    const clipboardItems = await navigator.clipboard.read();

    for (const clipboardItem of clipboardItems) {
      for (const type of clipboardItem.types) {
        const blob = await clipboardItem.getType(type);
        // we can now use blob here
        console.log(blob);
      }
    }

    //}
  } catch (e) {
    console.error(e);
  }
}
```

at `console` on this page using basic MDN example and got this error

```
NotAllowedError Type application/octet-stream not supported on write.
```

so appears a few things need to change for this to work as I expected. 

I already proposed this https://github.com/guest271314/requestClient which would solve https://github.com/GoogleChrome/chrome-extensions-samples/issues/766, filed as an issue here in https://github.com/GoogleChrome/chrome-extensions-samples/issues/781 - explained in the repository above. 

So, I think I have laid out several solutions for #766 already - including setting to `opener` of the offscreen document https://github.com/GoogleChrome/chrome-extensions-samples/issues/778 which was previous possible with `setSelfAsOpener` in MV2 that no longer works. Ideally though we would not load an additional document at all - just dual-opt in client and `ServiceWorker`. 

I tried to file a FUGU, however a few stakeholders have permanenetly banned me. I was able to file this a few minutes ago https://github.com/wintercg/proposal-common-minimum-api/issues/36.

> Banned from Chromium and GitHub repos of the W3C? Ooops. Then I guess you're out of luck.
>  
> 
> On Mon, Jan 9, 2023 at 3:31 AM Thomas Steiner <[tomac@google.com](mailto:tomac@google.com)> wrote:
> Hi there,
> 
> This sounds more like an (extension) service worker feature request, so the venues would be the [service worker repo](https://github.com/w3c/ServiceWorker) or the [webextensions repo](https://github.com/w3c/webextensions). You can open new issues there. 
> 
> Cheers,
> Tom

I've filed a PR, and composed two solutions for #766 (https://github.com/guest271314/sw-transfer-stream; https://github.com/guest271314/offscreen-webrtc). There is also this https://github.com/guest271314/native-messaging-quickjs/tree/dynamic-externally-connectable to dynamically set externall connectable - though we are still using JSON for communication.

When you find the time kind review the above issues and proposals. 'Bout all a guy can do at this point re #766.

Carry on.

jpmedley (changes requested):
Simeon,

In addition to these changes, do you want to write a README? We talked about doing this in the docs/samples weekly, last week. Having them all would be ideal, but this seems like a case where adding them as we go is better than not having them. Here's [the requirements I recorded](https://github.com/GoogleChrome/chrome-extensions-samples/issues/798) last week.

cookbook/offscreen-clipboard-write/background.js
```diff
@@ -0,0 +1,56 @@
...
+chrome.action.onClicked.addListener(async () => {
+  await addToClipboard(textToCopy);
+});
+
+// Solution 1 - As of Jan 2023, service workers cannot directly interact with the system clipboard.
+// To work around this, we'll create an offscreen document and pass it the data we want to write to
+// the clipboard.
+async function addToClipboard(value) {
+  // This pattern show in this if-else ensures that the offscreen document exists before we try to
+  // send it a message. If we didn't await the `hasDocument()` and `createDocument()` calls, the
```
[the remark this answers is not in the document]
dotproto: I've likely done this in other samples, but I'm not 100% sure. I would like to make inline symbol decoration a standard practice for our code samples. My rationale is outlined in [this comment](https://github.com/GoogleChrome/chrome-extensions-samples/pull/797/files#r1069986899).

cookbook/offscreen-clipboard-write/background.js
```diff
@@ -0,0 +1,56 @@
...
+  // Now that are sure we have an off document, we can safely dispatch the message.
+  chrome.runtime.sendMessage({
+    type: 'copy-data-to-clipboard',
+    target: 'offscreen-doc',
+    data: value,
+  });
+}
+
+// Solution 2 – Once extension service workers can use the Clipboard API,
+// replace the offscreen document based implementation with something like this.
```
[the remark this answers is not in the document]
dotproto: We don't have the work for exposing the Clipboard API in extension service workers scheduled but it is (loosely) something we plan to do. 

* [crbug 1160302](https://crbug.com/1160302) tracks a request to expose clipboard functionality in the extension service worker. 
* [crbug 1404835](https://crbug.com/1404835) specifically requests that we relax the Clipboard API's focus requirement for extensions. 

I'm not strongly opposed to removing this comment and (unused) function, but I would like to explain my rationale for including it before doing so. My goal in including it here was to reinforce the idea that developers should plan for the future when leveraging offscreen documents. Invoking an offscreen document requires that the extension provide one or more reasons (enum) and a justification (string) to identify why the extension is using an offscreen document in the first place. This information informs Chrome what kind of lifetime enforcement it should use for the offscreen document, helps provide reviewers with context on why and how an offscreen document is being used, and finally may be used to inform the user when abnormal behavior is detected.

Once a given web platform feature is exposed in extension service workers, developers should migrate the relevant code out of the offscreen document and into the extension service worker. This will allow the extension to perform the same work faster and using less resources. Developers that are not architecting their code with this future migration in mind may end up having to re-write their implementation when this change eventually comes.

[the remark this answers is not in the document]
IanStanion-google: @jpmedley How does this context on the idea affect your position? 
If you are opposed to this kind of future-proofing of the sample code itself, do you think it might be useful to add some information related to it into the readme, or should we keep the code sample focused on simply showcasing this feature?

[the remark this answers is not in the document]
jpmedley: It's not really resolved, but we're punting until we can have a policy discussion about how we discuss future plans.

cookbook/offscreen-clipboard-write/offscreen.js
```diff
@@ -0,0 +1,62 @@
...
+
+  // `document.execCommand('copy')` works against the user's selected in a web page. As such, we
+  // need to insert the updated text into the document and select it before it can be added to the
+  // clipboard.
+  textEl.value = data;
+  textEl.select();
+  document.execCommand('copy');
+}
+
+// This textarea is used to easily modify and select it's text value.
```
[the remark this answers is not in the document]
jpmedley: This is still unaddressed.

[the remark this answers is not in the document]
dotproto: This may be too much of an implementation detail related to working with `document.execCommand('copy')`. Since this function works against a user's selection in a document, we need to programmatically select the text we want to add to the clipboard before calling execCommand. It's possible to select arbitrary text content in a page using [Range](https://developer.mozilla.org/en-US/docs/Web/API/Range/Range), but form controls like `<input>` or `<textarea>` provide a `select()` method that will immediately set the selection range on the element's content. We prefer the use of `<textarea>` over `<input>` because the former supports multiline content out of the box; using `<input>` here will strip newlines while `<textarea>` will preserve the plain text formatting of our input string.

dotproto:
@jpmedley, PR has been updated to address recent feedback. Should be ready for another pass.

jpmedley (approved):

jpmedley:
PR submitter does the merge.

laplus-sadness:
Hello. Thank you for finally writing an example using the offscreen api, but could someone please update the [documentation](https://developer.chrome.com/docs/extensions/reference/offscreen/) with the function `hasDocument` and the result of all the methods possibly being async?
The documentation specifies that returning a promise is still pending to be added to chrome.

tophf:
All methods already return a Promise.
The `Pending` label means this method is not yet released in the stable Chrome.

laplus-sadness:
> All methods already return a Promise.

Thank you for taking the Pending label out @tophf, but does all the definitions return a Promise?

> In Manifest V3 and later, return a `Promise` by omitting the `callback` argument. The type inside the `Promise` is the same as the 1st argument to `callback`.

Because the documentation says what's quoted above, I was under the impression that there were two overloading possibilities for each method:

```typescript
function createDocument(parameters: CreateParameters, callback: () => void): void;
function createDocument(parameters: CreateParameters): Promise<void>;
```

Instead of:

```typescript
function createDocument(parameters: CreateParameters, callback?: () => void): Promise<void>;
```

Which of them is right?

tophf:
@laplus-sadness, as I explained above the label "Pending" has nothing to do with the returned type. It's only about availability in the stable Chrome.

tophf:
BTW, here's the source that shows the actual state: https://crsrc.org/extensions/common/api/offscreen.idl. The same exists for every other API, although some of them use json not idl.

laplus-sadness:
Thank you. Sorry for bothering you with all this

tophf:
No problem. I also advertised the source code, so hopefully it'll be useful too. BTW this API documentation says `109+` so it no longer contains the confusing Pending labels.

PR #801, opened 2023-01-08
Update manifest.json with minor corrections to description.

SmashedFrenzy16:
Update `manifest.json` with minor corrections to description.


jpmedley:
Thank you for your submission. 'Documentation articles' is not a proper noun and therefore should not be capitalized.

PR #806, opened 2023-01-10
Split favicon into two separate examples

AmySteam:
Creates two examples of how to use the favicon API on a popup (extension page) and on a content script


examples/favicon-cs/content.js
```diff
@@ -0,0 +1,14 @@
+console.log("content-script")
```
oliverdunk, line 1: Should we remove this before merging?

examples/favicon-cs/README.md
```diff
@@ -0,0 +1,16 @@
...
+See [Fetching favicons](https://developer.chrome.com/docs/extensions/mv3/favicon) to learn more. 
+
+## Testing the extension
+
+1. Follow the instructions to load an [unpacked extension](https://developer.chrome.com/docs/extensions/mv3/getstarted/development-basics/#load-unpacked).
+2. Navigate to [www.example.com](https://www.example.com/).
+
+It should look like this:
+
+<img src="https://wd.imgix.net/image/BhuKGJaIeLNPW9ehns59NfwqKxF2/3Q1glvnzbWhraXRtnGOy.png" alt="Content script using the Favicon API" width="400"/>
```
oliverdunk, line 16: It looks like this is our first time using imgix, are we comfortable hosting there?

oliverdunk (changes requested):
Tested both examples and they're working great, just left a couple of nits :)

examples/favicon-cs/content.js
AmySteam: Yep! Missed that 😅

examples/favicon-cs/README.md
AmySteam: Oh right, I forgot to mention this is where all images of https://developer.chrome.com/ are hosted.
I don't think this is a blocker, but it's probably something we should agree on with the team. Thanks for bringing this up 👍

examples/favicon-cs/manifest.json
```diff
@@ -0,0 +1,20 @@
...
+      "js": ["content.js"],
+      "css": ["style.css"]
+    }
+  ],
+  "permissions": ["favicon"],
+  "web_accessible_resources": [
+    {
+      "resources": ["_favicon/*"],
+      "matches": ["<all_urls>"],
+      "extension_ids": ["*"]
```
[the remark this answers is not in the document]
AmySteam: Will do!

oliverdunk (approved):
LGTM 👍

PR #829, opened 2023-02-16
Migrate chrome.printing api sample from v2 to v3 manifest

rayxu-google:
1. move chrome.printing api sample out of _archive
2. migrate it's manifest from v2 to v3


oliverdunk (approved):
Thanks for this! I added a commit restoring the MV2 sample since I think it's nice to have both.

PR #830, opened 2023-02-21
Add API sample for chrome.scripting API.

oliverdunk:
Adds an API sample for the `chrome.scripting` API. This provides a basic way to register content scripts dynamically or programmatically execute a script and experiment with a range of options.

image


api-samples/scripting/sw.js
```diff
@@ -0,0 +1,29 @@
...
+// https://developers.google.com/open-source/licenses/bsd
+
+chrome.action.onClicked.addListener(openDemoTab);
+
+function openDemoTab() {
+  chrome.tabs.create({ url: 'index.html' });
+}
+
+chrome.webNavigation.onDOMContentLoaded.addListener(async ({ tabId, url }) => {
+  if (url === "https://example.com/#inject-programmatic") {
```
sebastianbenz, line 14: Suggestion: use an early exit to avoid indentation

api-samples/scripting/README.md
```diff
@@ -0,0 +1,24 @@
+# chrome.scripting API
+
+This sample demonstrates basic usage of the [chrome.scripting](https://developer.chrome.com/docs/extensions/reference/scripting/) API.
```
sebastianbenz, line 3: suggestion rephrase to make it more clear what this example does (for people like me who don't know what the scripting API does...)
```suggestion
This sample demonstrates how to use the [chrome.scripting](https://developer.chrome.com/docs/extensions/reference/scripting/) API to inject JavaScript into web pages.
```

sebastianbenz (commented):
This is great! Left a few nits. We should also use the Apache 2.0 license.

Sidenote: imo this is not really a sample, but a demo.

oliverdunk:
@sebastianbenz This should be good for another look! Thanks for the feedback :)

sebastianbenz (approved):
LGTM - thanks! One note: the tab UI is a bit confusing as the selected tab label has a different label than the tab content.

oliverdunk:
Inverted the colours so I think we're good to go! Merging.

PR #841, opened 2023-03-03
Add wasm sample

### LINKED ISSUE #775: Sample: Using WASM in Manifest V3
I'd like to try working on a PR to demonstrate using WASM in MV3. Before doing so, I wanted to propose it here. I've tried my best to follow the format which @dotproto used in other issues (e.g https://github.com/GoogleChrome/chrome-extensions-samples/issues/766).

**Overview**

A sample showing how to load WASM in Manifest V3. This requires the new `wasm-unsafe-eval` CSP directive (see https://developer.chrome.com/docs/extensions/mv3/mv3-migration/#content-security-policy).

**Proposed Implementation**

I think a nice demo would be a simple extension which prints "Hello World!" from Rust.

- Manifest with `wasm-unsafe-eval`
- lib.rs file that prints "Hello World" in the start function
- Service worker script which loads the WASM, using `importScripts` to include the generated JS

**Additional Thoughts**

Deciding where to put this is a little hard. Looking at existing discussions ([here](https://github.com/GoogleChrome/chrome-extensions-samples/pull/746#pullrequestreview-1092260413)), I suspect `cookbook` may be the best place as this feels beyond an API demo but is not a fully functioning extension. We may want to commit the compiled WASM to avoid the need for a build step, or alternatively that could be documented in the README.

I also considered demonstrating how to access something like `self.clients` since this is not covered well in general WASM tutorials, which seem to focus heavily on webpages. However, in the interests of keeping this demo simple I suspect that may be something to revisit in the future.

daidr:
Adds a sample for using wasm in Manifest V3, I kept the wasm-pack build artifacts but I'm not sure if that's right.

fixes #775


functional-samples/cookbook.wasm-helloworld-print-nomodule/README.md
```diff
@@ -0,0 +1,42 @@
...
+   ```
+
+3. Build WASM.
+
+   ```bash
+   cd wasm
+   wasm-pack build --target no-modules
+   ```
+
+[0]: https://developer.chrome.com/docs/extensions/mv3/mv3-migration/#content-security-policy
```
oliverdunk, line 41: It looks like this page has moved. Perhaps we could link to https://developer.chrome.com/docs/extensions/mv3/manifest/content_security_policy/? That's also nice because it isn't in the migration section so it is more "evergreen".

functional-samples/cookbook.wasm-helloworld-print-nomodule/manifest.json
```diff
@@ -0,0 +1,12 @@
...
+  "version": "1.0",
+  "manifest_version": 3,
+  "description": "Inspect background service worker console for output",
+  "background": {
+    "service_worker": "background.js"
+  },
+  "content_security_policy": {
+    "extension_pages": "default-src 'self' 'wasm-unsafe-eval'"
+  }
+}
```
oliverdunk, line 12: Could you make sure to add newlines to each file?

functional-samples/cookbook.wasm-helloworld-print-nomodule/wasm/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+
+#[wasm_bindgen]
+pub fn print() {
+    console::log_1(&"[from wasm] Hello World!".into());
+}
+
+#[wasm_bindgen]
+pub fn print_with_value(value: &str) {
+    // with format! macro
+    console::log_1(&format!("[from wasm] Hello {}", value).into());
```
oliverdunk, line 21: We can probably remove this line? The point of this tutorial is to teach using WASM in extensions, not Rust, so I don't think we need to show `format!`.

functional-samples/cookbook.wasm-helloworld-print-nomodule/wasm/src/lib.rs
```diff
@@ -0,0 +1,25 @@
...
+
+    // with 2-args log function
+    console::log_2(&"[from wasm] Hello ".into(), &value.into());
```
oliverdunk, line 24: The browser adds a space automatically so this ends up with two spaces:

```
[from wasm] Hello  John
```

Can we remove the space to avoid that?

functional-samples/cookbook.wasm-helloworld-print-nomodule/wasm/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+extern crate wasm_bindgen;
```
oliverdunk, line 1: I think we can remove this line: https://doc.rust-lang.org/edition-guide/rust-2018/path-changes.html#no-more-extern-crate

functional-samples/cookbook.wasm-helloworld-print/README.md
```diff
@@ -0,0 +1,57 @@
+# WASM HelloWorld Demo
```
oliverdunk, line 1: Similar note on updating to use the README template.

functional-samples/cookbook.wasm-helloworld-print/README.md
```diff
@@ -0,0 +1,57 @@
...
+
+This recipe shows how to use WASM in Manifest V3.
+
+To load WASM in Manifest V3, we need to use the `wasm-unsafe-eval` CSP directive ([Content Security Policy][0]).
+
+## Running this extension
+
+1. Clone this repository.
+2. Load this directory in Chrome as an [unpacked extension][2].
+3. Find the extension named "WASM Load Example - Helloworld" and open console panel of the background service worker.
```
oliverdunk, line 11:
```suggestion
3. Find the extension named "WASM Load Example - Helloworld" and inspect the service worker.
```

functional-samples/cookbook.wasm-helloworld-print/README.md
```diff
@@ -0,0 +1,57 @@
...
+ // manifest.json
+ ...
+ "background": {
+     "service_worker": "background.js",
++    "type": "module"
+ },
+ ...
+```
+
+[0]: https://developer.chrome.com/docs/extensions/mv3/mv3-migration/#content-security-policy
```
oliverdunk, line 55: These two links need updating, similar to above :)

oliverdunk (changes requested):
I left a few comments and some suggested changes, but overall this looks great! I think keeping the generated artifacts is the right thing since it keeps this simple for people to try. Thanks for working on it!

oliverdunk (approved):
Amazing work on this, super excited to have the new samples! LGTM 🚀

patrickkettner:
thanks so much, @daidr! really stoked to see this land!

PR #842, opened 2023-03-06
Update mv3 native messaging

### LINKED ISSUE #644: Update Native Messaging API sample to MV3
We are trying to build a program that would be able to communicate with extensions on a remote computer without having to install something. 

However, it looks like the Chrome API has been updated and thus the functions we intended to use are now obsolete. Is there any way for us to find similar functions?

Thanks in advance!

Shubham-Rasal:
The Native Messaging Sample with mv2 is updated to mv3 in this sample.
 
Here is the updated sample's screeshots and it works as intended.

image

Fixes #644


_archive/mv2/api/nativeMessaging/host/native-messaging-example-host
```diff
@@ -1,19 +1,14 @@
-#!/usr/bin/env python
-# Copyright (c) 2012 The Chromium Authors. All rights reserved.
```
helfrichmichael, line 2: For this portion, keep the license and adjust the env to `#!/usr/bin/python3` to ensure that Chrome launches  this file with Python3, otherwise it will likely fail as it doesn't know what the file should be ran as.

_archive/mv2/api/nativeMessaging/host/install_host.sh
```diff
@@ -6,14 +6,14 @@
...
-  if [ "$(whoami)" = "root" ]; then
+if [ $(uname -s) == 'Darwin' ]; then
+  if [ "$(whoami)" == "root" ]; then
     TARGET_DIR="/Library/Google/Chrome/NativeMessagingHosts"
   else
     TARGET_DIR="$HOME/Library/Application Support/Google/Chrome/NativeMessagingHosts"
   fi
 else
-  if [ "$(whoami)" = "root" ]; then
+  if [ "$(whoami)" == "root" ]; then
```
helfrichmichael, line 16: Under the root instance to ensure that the script is accessible by all users, set the `native-messaging-example-host` file to execute.

`chmod a+x "$DIR/native-messaging-example-host"`

api-samples/nativeMessaging/app/manifest.json
```diff
@@ -0,0 +1,20 @@
+{
+  "manifest_version": 3,
+  "key": "MIGfMA0GCSqGSIb3DQEBAQUAA4GNADCBiQKBgQDcBHwzDvyBQ6bDppkIs9MP4ksKqCMyXQ/A52JivHZKh4YO/9vJsT3oaYhSpDCE9RPocOEQvwsHsFReW2nUEc6OLLyoCFFxIb7KkLGsmfakkut/fFdNJYh0xOTbSN8YvLWcqph09XAY2Y/f0AL7vfO1cuCqtkMt8hFrBGWxDdf9CQIDAQAB",
+  "name": "Native Messaging Example",
+  "version": "1.0",
+  "description": "Send a message to a native application.",
+  "background": {
```
helfrichmichael, line 7: main.js relies on `document.*` which isn't available in service workers. Since main.html launches as a popup, this can be used instead of any background/service worker instance so you can just remove this background portion.

helfrichmichael (changes requested):
Thanks for putting this together. I had a chance to test it out and it seems to work properly with the following adjustments.

_archive/mv2/api/nativeMessaging/app/manifest.json
```diff
@@ -16,4 +16,4 @@
   "permissions": [
     "nativeMessaging"
   ]
-}
+}
```
oliverdunk, line 19: Could we make sure all files have newlines?

oliverdunk (changes requested):
Thanks @helfrichmichael for the review! @Shubham-Rasal, do you think you might be able to look at this feedback in the next couple of days? If not, no worries at all - I can make the changes and we can get this merged. Thanks so much for your work on it.

Shubham-Rasal:
> Thanks @helfrichmichael for the review! @Shubham-Rasal, do you think you might be able to look at this feedback in the next couple of days? If not, no worries at all - I can make the changes and we can get this merged. Thanks so much for your work on it.

Sorry Oliver, I won't be able to work on this. Please understand.

oliverdunk (approved):
Pushed some changes and I think this is good to merge. Thanks for the work on this.

PR #844, opened 2023-03-08
Update downloads to mv3

Shubham-Rasal:
This PR aims to upgrade the existing download extension samples to MV3.  Here I have migrated the downloads manager and download links samples. 
If this PR is appropriate, I would love to continue to update others as well


patrickkettner:
@oliverdunk would you be able to review?

api-samples/downloads/download_links/send_links.js
```diff
@@ -0,0 +1,40 @@
...
+  if (((i > 0) && (links[i] == links[i - 1])) ||
+      (links[i] == '') ||
+      (kBadPrefix == links[i].toLowerCase().substr(0, kBadPrefix.length))) {
+    links.splice(i, 1);
+  } else {
+    ++i;
+  }
+}
+
+console.log(links)
```
oliverdunk, line 35: We weren't previously logging here, so I think we should probably remove this?

api-samples/downloads/download_links/manifest.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "Download Selected Links",
+  "description": "Select links on a page and download them.",
+  "version": "0.3",
+  "minimum_chrome_version": "16.0.884",
+  "permissions": [
+    "downloads",
+    "activeTab",
```
oliverdunk, line 8: Thanks for moving to activeTab here!

api-samples/downloads/download_manager/background.js
```diff
@@ -0,0 +1,267 @@
...
+  } else if (options.anyRecentlyCompleted) {
+    drawCompleteBadge(ctx);
+  }
+  return canvas;
+}
+
+function maybeOpen(id) {
+  var openWhenComplete = [];
+  try {
+    openWhenComplete = JSON.parse(localStorage.openWhenComplete);
```
oliverdunk, line 143: localStorage isn't available in MV3. We'll need to update this before we can test the extension.

oliverdunk (changes requested):
Thanks so much for taking a look at this!

I left some initial thoughts - in both cases the extensions aren't working right now so it would be great to fix that. Maybe you had some extra changes that you didn't commit?

It would also be great to run the linter (`npm run lint:fix`). Note that this will change `var` to `let` which will break the `download_links` extension when you open the popup more than once, since it will try to re-declare variables. We'll need to look at fixing that.

Shubham-Rasal:
Thanks for the feedback. On it.

jpmedley:
Does either of you have time to add READMEs to these?

Shubham-Rasal:
Yeah I will add them, once the code is ready.

Shubham-Rasal:
Sorry, I won't be able to work on this. You can go ahead with your plan

PR #847, opened 2023-03-10
adding print job table along with cancel button

rayxu-google:


jpmedley:
This does a few things I don't think we should do. Unless I'm misreading, it makes the 'Cancel' button visible at all times even though `cancelJob()` doesn't work at all times. It also puts the job status and 'Cancel' button at the bottom of the popup. If I'm trying this sample, and I need to scroll to get to the 'Cancel' button, I might not be able to it before the print job starts. (I might not want to actually print just to test a sample.)

rayxu-google:
I added more changes:
1. Cancel button will be hidden for job status not equal to "PENDING" and "IN_PROGRESS".
2. Scroll to the bottom of extension popup when print button is clicked. The Chrome extension has maximum size, it is difficult to display a table of printers and another table of print jobs in such small size of window, and we cannot predict the position of print button if user has a lot of printers in their table, after click on the print button, automatically scroll down to the bottom of window will let user be able to see the print job table and they can see the job status and cancel button.

Issue #807, opened 2023-01-11
Sample: Calling element.click() on an element with an inline event handler


dotproto:
## Overview

Starting with Manifest V3, Chrome introduced a strict content security policy (CSP) for content scripts run in the extension's isolated world (default behavior). This CSP does not allow the use of `unsafe-inline`. This prevents extensions from appending `<script>` tags with arbitrary payloads to a document, but it _also_ prevents other extensions from calling the `click()` method on elements that have click handlers.

This issue request the creation of an extension sample that demonstrates how to click an element that has an inline event  while still complying with the Chrome Web Store's policies forbidding "remotely hosted code" (executing code that is not bundled in the extension's package). 

## Proposed implementation

If a website contains inline event handlers, the page's CSP must allow `unsafe-inline` or those event handlers won't work. As such, we can work around extension CSP restrictions by performing the `click()` call inside a main world content script.

The user experience for this implementation is roughly as follows:

1. The first time the user clicks the extension's action button, a new tab will open to a demo site with inline event handlers. 
2. When the user clicks the extensions' action button while the demo site is in the currently focused tab, the extension will perform a "click()" on an element with an inline event handler. 

The extension will have 3 primary components: `background.js`; `content-isolated.js`, which will be injected into the extension's isolated world; and `content-main.js` (or equivalent function in background.js), which will be injected into the page's main world.

`content-main.js` will perform the click on behalf of the extension. It will receive a message from `content-isolated.js` 

In order to communicate between the main world and isolated world content scripts, the extension will dispatch a custom event that uses a nonce name known to both the main and isolated worlds. This pattern helps ensure that the event's cross-world communication does not interfere with normal page operations. 

## Additional considerations

The README for this sample should highlight that the use of custom events with nonce names does not guarantee that a hostile page will not intercept the message, as the page could monkeypatch the JavaScript environment in a way that allows it to intercept object constructors or event listener registrations. 

CustomElements can pass data across the world boundary by setting the `detail` property when instantiating a new custom element, but only if that data is cloneable with the structuredClone algorithm. If it cannot be cloned, the custom element's event.detail property in the receiving world will be `null`.

Issue #769, opened 2022-10-31
Demo request: Calling .click() on an element with an inline event handler or link with a javascript: scheme


dotproto:
**Summary**

Manifest V3's CSP for content scripts prevents extensions from executing line JavaScript. Typically inline script execution would occur when an element has inline event handlers bound or when an anchor tag's href attribute contains a `javascript:` scheme. As a result, if an extension tries to call `.click()` on such an element inside an isolated world content script, Chrome will throw a CSP error.

This demo should show at least one safe way that developers can work around this issue. 

**Suggested implementation**

Create a main world script (proxy-click.js?) that performs the click in the main world on behalf of the isolated script. At the moment I see a couple different ways we could approach this.

1. Inject both a main and isolated world script as functions with a nonce as the args. Use this to dispatch a custom event with the nonce in the name in order to prevent other scripts from seeing the messages passed between main and isolated worlds. 
2. Inject a main world script a document start, append an event listener to window, & listen for something like 'proxy-click'. When dispatching this event from the isolated world, emit as a capture event and in the handler use `stopPropagation()` to minimize visibility.
3. On demand, inject a main world script function with a CSS selector arg to perform the click. 

Of these, at the moment (1) seems best.

**Additional context**

*  [Issue 1299742](https://bugs.chromium.org/p/chromium/issues/detail?id=1299742): MV3 content script can't click a web page link with a javascript: URL

tophf:
> CSS selector arg

A better solution is to send the node itself, which will also allow clicking inside closed shadow roots:

```js
const elem = document.querySelector('a[onclick^="javascript:"]');
window.dispatchEvent(new MouseEvent(eventId, {relatedTarget: elem}));
```

guest271314:
A rather elaborate workaround for this would be to get the text in `console.log()` set the inline event handler, then set the `href` to an empty string, define an event handler with `addEventListener()` and the handler including `console.log()` with the extracted string passed, then call `click()`.

Another elaborate option is to again try to parse the existing JavaScript after `javascript:`, keep a .wat template which will be modified, then use `WebAssembly.instantiate()` and `env` to print the text.

blairjo:
This issue has been open for a while, although I can see a new sample was pushed by @daidr but never accepted: https://github.com/GoogleChrome/chrome-extensions-samples/pull/849

Is this a viable workaround to the issue of being able to click an element with inline code?

[...]

PR #849, opened 2023-03-15
Add a sample of clicking an anchor element with the javascript: scheme

daidr:
Relevant issues: #807 #769


oliverdunk:
@daidr, based on our email discussion would it be reasonable to call this a draft for now? Or would you like a review?

daidr:
> @daidr, thank you for your hard work on this. So that you're not shocked, I want to let you know we will replace your instructions for unpacked loading with [a link to ours](https://developer.chrome.com/docs/extensions/mv3/getstarted/development-basics/#load-unpacked). Something like this should theoretically included in every README. The problem is that if they ever change the UI for loading unpacked we'll have a maintenance problem on our hands.

I understand. What do I need to do?

Dandy0812 (approved):

patrickkettner:
@oliverdunk this one stalled on your call between landing it as is or asking for the relatedTarget rework, any thoughts?

oliverdunk:
I'm going to close this. As always, I really appreciate the work, but I think this is more of a workaround than something we would want to encourage unless absolutely essential. I also haven't seen this come up much recently, so I don't think there are a lot of developers having issues with it.

PR #848, opened 2023-03-14
[New sample] Open Chrome API reference page (omnibox, alarms, messaging sample)

AmySteam:
## User Story
As a web developer, I want to open extension API reference pages on Chrome, so that I can quickly learn how to use a specific extension API.

**Bonus**: A daily extension tip will be displayed as a popover after clicking the "tip" navigation list item.
(Currently, it requires Canary build and the "Experimental Web Platform features" flag enabled. The release date was moved to M114, so this approach may need to be abandoned if we want to launch before then.)

### How to use it?
1. Enter the keyword "api" in the browser address bar.
2. Then, press "tab" or "space".
3. Start typing they API keyword.
4. A list the most common APIs will show or a list of past searches
5. Select the API desired, and a new page will open to the Chrome API reference page.

### TODO:
- [ ]  The [Popover API](https://chromestatus.com/feature/5463833265045504) is currently being developed. If it's not ready by the launch, then we will make adjustments. The popover was chosen to simplify the example.
- [ ] The ReadME.md will have a link to the tutorial

PR #851, opened 2023-03-17
[New tutorial] Open Chrome API docs (condensed) 

AmySteam:
Initial [PR#848](https://github.com/GoogleChrome/chrome-extensions-samples/pull/848/) fits more into a robust/useful developer tool example. So It will be moved to a different folder. This is a simpler approach with less omnibox suggestions.

---
## User Story
As a web developer, I want to open extension API reference pages on Chrome, so that I can quickly learn how to use a specific extension API.

**Bonus**: A daily extension tip will be displayed as a popover after clicking the "tip" navigation list item.
(Currently, it requires Canary build and the "Experimental Web Platform features" flag enabled. The release date was moved to M114, so this approach may need to be abandoned if we want to launch before then.)

### How to use it?
1. Enter the keyword "api" in the browser address bar.
2. Then, press "tab" or "space".
3. Enter the complete name of the API.
4. OR choose from a list of past searches
5. Select the API desired, and a new page will open to the Chrome API reference page.


functional-samples/tutorial.open-api-reference/content.js
```diff
@@ -0,0 +1,30 @@
+// Popover API https://chromestatus.com/feature/5463833265045504
+
+(async () => {
+  const nav = document.querySelector('.navigation-rail__links');
+
+  const { tip } = await chrome.runtime.sendMessage({ greeting: 'tip' });
+
+  const tipWidget = createDomElement(`
```
sebastianbenz, line 8: innerHtml instead of append would be shorter - would that be OK?

functional-samples/tutorial.open-api-reference/sw-omnibox.js
```diff
@@ -0,0 +1,40 @@
...
+  chrome.tabs.create({ url: URL_CHROME_EXTENSIONS_DOC + input });
+  // Save the latest keyword
+  updateHistory(input);
+});
+
+async function updateHistory(input) {
+  const { apiSuggestions } = await chrome.storage.local.get('apiSuggestions');
+  apiSuggestions.unshift(input);
+  apiSuggestions.splice(NUMBER_OF_PREVIOUS_SEARCHES);
+  await chrome.storage.local.set({ apiSuggestions });
```
sebastianbenz, line 39: Remove the await here and return the promise from local.set instead. Callers than can decide whether to await or ignore the promise.

functional-samples/tutorial.open-api-reference/content.js
AmySteam: Can you give me an example? 

If we want even less code, I could make the button less fancy:

Screenshot 2023-03-21 at 15 47 26

functional-samples/tutorial.open-api-reference/sw-omnibox.js
```diff
@@ -0,0 +1,40 @@
...
+const NUMBER_OF_PREVIOUS_SEARCHES = 4;
+
+// Display the suggestions after user starts typing
+chrome.omnibox.onInputChanged.addListener(async (input, suggest) => {
+  await chrome.omnibox.setDefaultSuggestion({
+    description: 'Enter a Chrome API or choose from past searches'
+  });
+  const { apiSuggestions } = await chrome.storage.local.get('apiSuggestions');
+  const suggestions = apiSuggestions.map((api) => {
+    return { content: api, description: `Open chrome.${api} API` };
```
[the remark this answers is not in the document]
AmySteam: This happens in two places, so it would end up adding more code. Would you be opposed to leaving this as is?

functional-samples/tutorial.open-api-reference/sw-omnibox.js
AmySteam: Done

functional-samples/tutorial.open-api-reference/content.js
sebastianbenz: Sorry, missed that `nav` already contains existing elements. Your approach is better.

functional-samples/tutorial.open-api-reference/sw-omnibox.js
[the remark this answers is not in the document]
sebastianbenz: This feels like the wrong place, the method mixes event handling and data processing. A helper method which is used on install and updateHistory would be better imo.

Anyway, this is not really important for an example. Do whatever you think is best.

sebastianbenz (approved):

functional-samples/tutorial.open-api-reference/sw-omnibox.js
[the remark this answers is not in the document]
AmySteam: I'm leaving it as is for the sake of moving this project along, but we can always circle back :)

AmySteam:
@sebastianbenz Thanks for all your insights!!! 💯 😊

PR #853, opened 2023-03-20
Added Updated Calculator Sample

IanStanion-google:


sebastianbenz:
How much work would it be to remove JQuery (not sure we should continue recommending it) and migrate to modern JS?

oliverdunk:
As discussed offline, closing this one out. It doesn't demonstrate any extension concepts and is fairly outdated, so is probably not worth updating.

PR #854, opened 2023-03-20
eslint: validate script imports

sebastianbenz:
Add https://github.com/import-js/eslint-plugin-import for validing script imports (eslint will fail if imported file does not exist).

Also fixes prettier errors in api-samples.


sebastianbenz:
@patrickkettner this might be interesting for an extension specific linting setup, as the error messages you get in Chrome on a failed import are very cryptic.

AmySteam (approved):
Thanks @sebastianbenz! LGTM :)

PR #855, opened 2023-03-20
make google news samples functional

### LINKED ISSUE #533: 3 Chrome Extensions samples not working
Unfortunately the following Chrome Extensions samples: /extensions/news/, extensions/news_a11y/ and /extensions/news_i18n/ are not really working... 

They are full of errors including rss errors, content_security_policy errors and others...

patrickkettner:
fixes #533

there was a handful of CORs issues, and the google news API went from http to https, so that had to be updated.

Since they were broken, I figured we should either delete them or fix them, and the fixes were small.


jpmedley:
@patrickkettner, while you're here, do you have time to add a README using our new instructions? https://github.com/GoogleChrome/chrome-extensions-samples/blob/main/README-template.md

_archive/mv2/extensions/news/README
```diff
@@ -0,0 +1,7 @@
+# News
+
+This sample demonstrates a simple popup panel that access a remote URL. In this case, we are the Google News RSS feed, and building a UI to display that information.
```
jpmedley, line 3:
```suggestion
This sample demonstrates a popup panel that accesses a remote URL (in this case, a Google News RSS feed) and builds a UI to display that information.
```

_archive/mv2/extensions/news/README
```diff
@@ -0,0 +1,7 @@
...
+
+## Implementation Notes
+
+Note that this is a _very_ old example, and is kept here purely for archival purposes. It is not recommended to use this as a base of a new extension.
```
jpmedley, line 7:
```suggestion
Note that this is a _very_ old example, and is kept here purely for archival purposes. We do not recommend using this as a base of a new extension.
```

_archive/mv2/extensions/news/README
jpmedley: The corporate style guide likes us to avoid words like 'simple' and 'easy'. Removing it also applies a writing best practice to prefer saying things with fewer words when possible.

jpmedley (changes requested):

PR #856, opened 2023-03-20
Eslint: don't allow var

sebastianbenz:


functional-samples/sample.bookmarks/popup.js
```diff
@@ -51,7 +52,7 @@ function dumpNode(bookmarkNode, query) {
       chrome.tabs.create({ url: bookmarkNode.url });
     });
 
-    var span = $('<span>');
+    span = $('<span>');
```
patrickkettner, line 55: Why is this an empty string redefined? Is there a reason it can't just be a const?

sebastianbenz: scope

patrickkettner (approved):

functional-samples/sample.bookmarks/popup.js
sebastianbenz: yeah, const was also my first try. eslint is better at these things than us humans

patrickkettner: 🤖

PR #863, opened 2023-03-23
Added desktop capture sample

IanStanion-google:


sebastianbenz:
@oliverdunk could you please talk a look as it clearly falls into your realm :-)

sebastianbenz:
Just noticed that this is several month old. Is this still relevant or should we close it?

oliverdunk:
I spoke to Devlin and we weren't sure if there's feature parity in the web APIs yet or not. `cancelChooseDesktopMedia` is one that seems like a possible extension API only feature.

That said - I'm open to closing this given it's an API most people are unlikely to use. Will defer to you and Ian.

sebastianbenz:
I was wondering whether there's a UI difference for extensions vs web APIs, but couldn't get this sample to work (@IanStanion-google does it work for you?). If no, I think we should remove it.

IanStanion-google:
It doesn't work!
A long while ago me and Oliver agreed it would be alright to cut this one, I'll close it out now.

PR #865, opened 2023-03-23
[SW-tutorial] Refactor alarms API logic (sw-tips.js)

AmySteam:
After conducting some experiments, which included closing all my browser windows (something I never do), I discovered that the alarms logic was not working as intended. This was because alarms do not [persist across browser sessions](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/alarms#:~:text=Alarms%20do%20not%20persist%20across%20browser%20sessions.), which was frustrating 😖.

To remedy this issue, I've modified the code by giving the alarm a name and checking if an alarm with that name already exists. If it doesn't exist, the code will create the alarm again.

@sebastianbenz, could you assign me to update the Alarms API documentation? 🙏 ✨

### DIFF
--- a/functional-samples/tutorial.quick-api-reference/sw-tips.js
+++ b/functional-samples/tutorial.quick-api-reference/sw-tips.js
@@ -8,11 +8,14 @@ const updateTip = async () => {
   return chrome.storage.local.set({ tip: tips[randomIndex] });
 };
 
-// Creates an alarm if it doesn't exist
+const ALARM_NAME = 'tip';
+
+// Check if alarm exists to avoid resetting the timer.
+// The alarm might be removed when the browser session restarts.
 async function createAlarm() {
-  const alarm = await chrome.alarms.get('tip');
+  const alarm = await chrome.alarms.get(ALARM_NAME);
   if (typeof alarm === 'undefined') {
-    await chrome.alarms.create('tip', {
+    chrome.alarms.create(ALARM_NAME, {
       delayInMinutes: 1,
       periodInMinutes: 1440
     });
--- a/functional-samples/tutorial.quick-api-reference/sw-tips.js
+++ b/functional-samples/tutorial.quick-api-reference/sw-tips.js
@@ -8,13 +8,16 @@ const updateTip = async () => {
   await chrome.storage.local.set({ tip: tips[randomIndex] });
 };
 
-// Create a daily alarm and retrieves the first tip when extension is installed.
-chrome.runtime.onInstalled.addListener(({ reason }) => {
-  if (reason === 'install') {
-    chrome.alarms.create({ delayInMinutes: 1, periodInMinutes: 1440 });
+// Creates an alarm if it doesn't exist
+async function createAlarm() {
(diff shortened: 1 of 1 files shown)


functional-samples/tutorial.quick-api-reference/sw-tips.js
```diff
@@ -5,16 +5,22 @@ const updateTip = async () => {
...
+  return chrome.storage.local.set({ tip: tips[randomIndex] });
 };
 
-// Create a daily alarm and retrieves the first tip when extension is installed.
-chrome.runtime.onInstalled.addListener(({ reason }) => {
-  if (reason === 'install') {
-    chrome.alarms.create({ delayInMinutes: 1, periodInMinutes: 1440 });
+// Creates an alarm if it doesn't exist
+async function createAlarm() {
+  const alarm = await chrome.alarms.get('tip');
```
sebastianbenz, line 13: nit: extract alarm name into a constant

functional-samples/tutorial.quick-api-reference/sw-tips.js
```diff
@@ -5,16 +5,22 @@ const updateTip = async () => {
...
+  if (typeof alarm === 'undefined') {
+    await chrome.alarms.create('tip', {
```
sebastianbenz, line 15: Is the await necessary here?

functional-samples/tutorial.quick-api-reference/sw-tips.js
```diff
@@ -5,16 +5,22 @@ const updateTip = async () => {
...
   const randomIndex = Math.floor(Math.random() * tips.length);
-  await chrome.storage.local.set({ tip: tips[randomIndex] });
+  return chrome.storage.local.set({ tip: tips[randomIndex] });
 };
 
-// Create a daily alarm and retrieves the first tip when extension is installed.
-chrome.runtime.onInstalled.addListener(({ reason }) => {
-  if (reason === 'install') {
-    chrome.alarms.create({ delayInMinutes: 1, periodInMinutes: 1440 });
+// Creates an alarm if it doesn't exist
```
sebastianbenz, line 11: Can you extend the comment to explain why you're doing this

sebastianbenz (commented):
Good find! That's why we need to build samples 🙂 . Thinking about this, setAlarm is probably the wrong API to use here. Maybe [periodic background sync](https://developer.mozilla.org/en-US/docs/Web/API/Web_Periodic_Background_Synchronization_API) would be a better way to do it (if it works in a sw)

sebastianbenz (approved):

AmySteam:
@sebastianbenz I think switching to [periodic background sync](https://developer.mozilla.org/en-US/docs/Web/API/Web_Periodic_Background_Synchronization_API) is out of scope for the tutorial. But maybe can be included in the explainer?

sebastianbenz:
@AmySteam definitely, I'd also add a comment in the sample mentioning that setAlarm serves demonstration purposes, but periodic background sync would be a better approach.

PR #891, opened 2023-05-01
added search excludes and json schemas for manifest.json files

benjaminloerincz:


patrickkettner:
thanks! But we don't normally add editor specific files

PR #892, opened 2023-05-01
added length to avoid error

benjaminloerincz:


patrickkettner:
Hi @benjaminloerincz !
What error? `[].map` wouldn't error unless tabs in not an array, in which case checking `.length` would also error

benjaminloerincz:
Hi @patrickkettner,
without the check, I receive this error in dev tools if no matching tab is opened:
`Uncaught (in promise) TypeError: Error in invocation of tabs.group(object options, optional function callback): Error at parameter 'options': Error at property 'tabIds': Value did not match any choice.`
in this line
`const group = await chrome.tabs.group({ tabIds });`

patrickkettner:
ah, ok - thanks!
Rather than a length check, can you use optional chaining? i.e. `tabs?.group(...)`

benjaminloerincz:
I have tried your suggestion:
```
button.addEventListener('click', async () => {
  const tabIds = tabs.map(({ id }) => id);
  const group = await chrome.tabs?.group({ tabIds });
  await chrome.tabGroups.update(group, { title: 'DOCS' });
});
```
Unfortunately, the code throws this error on execution: `caught (in promise) TypeError: Error in invocation of tabs.group(object options, optional function callback): Error at parameter 'options': Error at property 'tabIds': Value did not match any choice.
    at HTMLButtonElement.<anonymous> (popup.js:49:36)`

patrickkettner (approved):
thanks!

PR #893, opened 2023-05-01
prefixed unused variables with underscore

benjaminloerincz:


patrickkettner:
Thanks for submitting the PR! This is not a style we use.

PR #896, opened 2023-05-01
moved body style to existing css file

benjaminloerincz:
* moved body style to existing css file
* added min-height: 150px; to make buttons visible if no bookmarks are available

### DIFF
--- a/functional-samples/sample.bookmarks/popup.css
+++ b/functional-samples/sample.bookmarks/popup.css
@@ -1,3 +1,8 @@
+body {
+  width: 400px;
+  min-height: 150px;
+}
+
 #editdialog input {
   width: 100%;
 }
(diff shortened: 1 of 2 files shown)


functional-samples/sample.bookmarks/popup.css
```diff
@@ -1,3 +1,8 @@
+body {
+  width: 400px;
+  min-height: 150px;
```
patrickkettner, line 3: why the additional rule?

benjaminloerincz: If you have no bookmarks, let's say you are using an new empty chrome profile, then the popup looks like this - which is fine:
image
If you try to add a bookmark to the bookmarks bar, you get this screen:
image
As you can see, the "Add" and "Cancel" button is not visible.

If you have one bookmark and try to edit the bookmark, you first see this - which is fine:
image
And after clicking "Edit", you will see this:
image
As you can see, the "Save" and "Cancel" button is missing.

The adjustment makes them visible:
Adding new bookmark:
image

Adjusting bookmark:
image

patrickkettner: thanks! your commit message didn't mention it, so it was unclear. appreciate the detailed response

patrickkettner (approved):

PR #903, opened 2023-05-01
corrected default_icon

benjaminloerincz:


patrickkettner (approved):
thanks! for repetitive clean up prs doing the same action (like adding images to manifests), feel free to bundle them into a single commit

PR #905, opened 2023-05-01
removed unused variables, cleanup

benjaminloerincz:


patrickkettner (approved):
thanks! for repetitive clean up prs doing the same action (like removing unused variables), feel free to bundle them into a single commit

PR #906, opened 2023-05-01
removed unused variables, switched to async/await, cleanup

benjaminloerincz:


api-samples/alarms/index.js
```diff
@@ -94,17 +93,12 @@ class AlarmManager {
...
-      chrome.alarms.clear(name, (wasCleared) => {
-        if (wasCleared) {
-          this.logMessage(`Manager: canceled alarm "${name}"`);
-        } else {
-          this.logMessage(`Manager: could not canceled alarm "${name}"`);
-        }
-
-        resolve(wasCleared);
-      });
+    await chrome.alarms.clear(name, (wasCleared) => {
```
sebastianbenz, line 96: Why await here and not a return? I think it'd be better to leave the decision whether the promise should be awaited to the caller.

api-samples/alarms/index.js
```diff
@@ -131,30 +125,21 @@ class AlarmManager {
...
-      chrome.alarms.clearAll((wasCleared) => {
-        if (wasCleared) {
-          this.logMessage(`Manager: canceled all alarms"`);
-        } else {
-          this.logMessage(`Manager: could not canceled all alarms`);
-        }
-
-        resolve(wasCleared);
-      });
+    await chrome.alarms.clearAll((wasCleared) => {
```
sebastianbenz, line 128: same as above

sebastianbenz (changes requested):

sebastianbenz (approved):
Thanks!

PR #913, opened 2023-05-13
Update README.md

mbaneshi:
spell checking issue
changed  "to" => "two " ,
with accordance to context


AmySteam (approved):
@mbaneshi good catch! 💯  Thanks for this PR 👍

PR #917, opened 2023-05-15
Rewrite handlebars-iframe example

Ambushfall:
#746 Re-write of original PR


oliverdunk:
Hey @Ambushfall! We have https://github.com/GoogleChrome/chrome-extensions-samples/pull/919 open at the moment which was opened just after yours (and has got slightly further through review, sorry about that). It looks like these are technically different samples but the functionality seems similar. Could you take a look and see if you think there is value in having both?

PR #936, opened 2023-05-31
Updated url_override samples to mv3 (minimal changes)

IanStanion-google:


bigleague767 (approved):

api-samples/override/override_igoogle/manifest.json
```diff
@@ -0,0 +1,9 @@
+{
```
oliverdunk, line 1: Thoughts on deleting this sample? It basically shows the same thing as the other one, and links to iGoogle which doesn't exist anymore, so I don't think it adds much.

patrickkettner: +1

jpmedley (approved):

PR #938, opened 2023-06-02
Devtools mv3 samples

IanStanion-google:


api-samples/devtools/panels/chrome-query/devtools.js
```diff
@@ -0,0 +1,28 @@
+// Copyright (c) 2012 The Chromium Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style license that can be
+// found in the LICENSE file.
```
oliverdunk, line 3: We should update this to the Apache license :)

api-samples/devtools/panels/chrome-query/devtools.js
```diff
@@ -0,0 +1,28 @@
...
+
+// The function below is executed in the context of the inspected page.
+/*global $0*/
+let page_getProperties = function () {
+  let data = window.jQuery && $0 ? jQuery.data($0) : {};
```
oliverdunk, line 8: We should probably do a pass updating the code style where we can (e.g, we could use const here).

oliverdunk (changes requested):
Thanks for this! Looks like a great start :)

oliverdunk (changes requested):
Thanks for this! Looks like a great start :)

IanStanion-google:
@oliverdunk Requested changes are in! Let me know if there are any other idioms in the code that should be different that I might have missed.

api-samples/devtools/chrome-query/chrome-query.md
```diff
@@ -0,0 +1,15 @@
+# devtools
+
+This sample demonstrates using the devtools API to add a sidebar that displays the jQuery data associated with the selected DOM element.
+
+## Overview
+
+The devtools window can be expanded with new elements created by the user. The extension contains html and js files that are injected into the devtools page to expand its functionality.
```
jpmedley, line 7:
```suggestion
This extension shows how DevTools can be expanded with new elements. It contains HTML and JavaScript files that are injected into the DevTools page to expand its functionality.
```

jpmedley (changes requested):

api-samples/devtools/chrome-query/chrome-query.md
```diff
@@ -0,0 +1,15 @@
+# devtools
```
sebastianbenz, line 1: Should this file be named readme.md?

api-samples/devtools/chrome-query/chrome-query.md
```diff
@@ -0,0 +1,15 @@
...
+
+This sample demonstrates using the devtools API to add a sidebar that displays the jQuery data associated with the selected DOM element.
```
sebastianbenz, line 3: Is it still worth demoing anything related to JQuery?

api-samples/devtools/panels/chrome-query/chrome-query.md
```diff
@@ -0,0 +1,7 @@
+# devtools
```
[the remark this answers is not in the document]
oliverdunk: Thanks for moving this, although can we rename it to panels? chrome-query doesn't make much sense IMO.

api-samples/devtools/chrome-query/chrome-query.md
oliverdunk: Long term I think a sample that doesn't use jQuery would be nice, but jQuery still exists, and so I don't think we necessarily need to change this given that our main goal is just having something that works in MV3. This feels like a useful sample for developers migrating even if the context is a bit less common these days.

api-samples/devtools/chrome-query/chrome-query.md
sebastianbenz: Sorry, my mistake, the naming convention is README.md, can you please update the name agian.

oliverdunk (approved):
LGTM 👍

PR #945, opened 2023-06-07
historyOverride sample for mv3

daidr:


oliverdunk (changes requested):
Took a quick look, I think as far as the migration goes this looks great. Would you also be up for spending a bit of time on a few improvements?

- We should probably hide pages where we don't have a title, since they just appear as blank right now. Running `URL.createObjectURL(new Blob())` and then opening the corresponding URL is an easy way to get one of those.
- Can we make the layout/design a bit nicer? Maybe just list the items in a table or at the very least we should remove the white background on the icon/text and just make the text a better color.

I don't want to overcomplicate this PR too much but I do think it looks a bit rough and ready right now and I'd rather tidy it up a bit before merging.

daidr:
> Took a quick look, I think as far as the migration goes this looks great. Would you also be up for spending a bit of time on a few improvements?
> 
> * We should probably hide pages where we don't have a title, since they just appear as blank right now. Running `URL.createObjectURL(new Blob())` and then opening the corresponding URL is an easy way to get one of those.
> * Can we make the layout/design a bit nicer? Maybe just list the items in a table or at the very least we should remove the white background on the icon/text and just make the text a better color.
> 
> I don't want to overcomplicate this PR too much but I do think it looks a bit rough and ready right now and I'd rather tidy it up a bit before merging.

I just updated the code and styles.

image

oliverdunk:
The new look is great! Thanks so much for that 😍 

I seem to be having trouble clicking the checkboxes. Do they work on your end?

daidr:
> The new look is great! Thanks so much for that 😍
> 
> I seem to be having trouble clicking the checkboxes. Do they work on your end?

Oh, this is a bug and it has been fixed now.

oliverdunk (approved):
LGTM! 👍

AmySteam (approved):
Thanks @daidr 💯

PR #947, opened 2023-06-08
Add storage sample

daidr:


api-samples/storage/stylizr/README.md
```diff
@@ -0,0 +1,13 @@
+# chrome.storage - Stylizr
+
+A sample that demonstrates how to use the [`chrome.storage`](https://developer.chrome.com/docs/extensions/reference/storage/) API.
+
+## Overview
+
+In this sample, the `chrome.storage` API is used to store the custom style string. The style will be applied to the active page when the extension's action icon is clicked.
```
jpmedley, line 7:
```suggestion
In this sample, the `chrome.storage` API stores a custom style string that will be applied to the active page when the extension's action icon is clicked.
```

jpmedley (changes requested):

jpmedley (approved):

api-samples/storage/stylizr/options.js
```diff
@@ -0,0 +1,54 @@
...
+  await storage.remove('css');
+  message('Reset stored CSS');
+  // Refresh the text area.
+  textarea.value = '';
+}
+
+function message(msg) {
+  const message = document.querySelector('.message');
+  message.innerText = msg;
+  setTimeout(function () {
```
oliverdunk, line 51: Can we add some logic to clear the timeout if this is called twice, so we don't end up running it early?

api-samples/storage/stylizr/popup.js
```diff
@@ -0,0 +1,35 @@
...
+        css: items.css,
+        target: {
+          tabId: currentTab.id
+        }
+      });
+    } catch (e) {
+      if (e) {
+        message.innerText = 'Not allowed to inject CSS into special page.';
+      } else {
+        message.innerText = 'Injected style!';
```
oliverdunk, line 25: I think this should be outside of the catch block, since that will only run if there's an error? It probably means you can remove the if(e) entirely.

api-samples/storage/stylizr/popup.js
```diff
@@ -0,0 +1,35 @@
...
+      }
+    }
+  } else {
+    const optionsUrl = chrome.extension.getURL('options.html');
+    message.innerHTML =
```
oliverdunk, line 30: Could we rewrite this to avoid using innerHTML? I want to encourage that wherever possible :)

oliverdunk (changes requested):
Left a few comments but overall this looks good. Thanks!

api-samples/storage/stylizr/popup.js
```diff
@@ -0,0 +1,36 @@
+// Store CSS data in the "local" storage area.
+const storage = chrome.storage.local;
+
+const message = document.querySelector('#message');
+
+// Check if there is CSS specified.
+(async function () {
```
oliverdunk, line 7: Can we give this function a name and then call it vs. using an IIFE?

api-samples/storage/stylizr/popup.js
```diff
@@ -0,0 +1,36 @@
...
+    try {
+      await chrome.scripting.insertCSS({
+        css: items.css,
+        target: {
+          tabId: currentTab.id
+        }
+      });
+      message.innerText = 'Injected style!';
+    } catch (e) {
+      message.innerText = 'Not allowed to inject CSS into special page.';
```
oliverdunk, line 23: Could we:

- Update the message to make it a bit less presumptuous. Maybe "Injection failed. Are you on a special page?"
- Log the error to the console as well

oliverdunk (changes requested):
Left a couple of final comments!

oliverdunk (approved):
LGTM! Thanks for the work on this one 🎉

PR #956, opened 2023-06-11
Privacy mv3 sample

IanStanion-google:
Fixes: https://github.com/GoogleChromeLabs/Extension-Docs/issues/85

### DIFF
--- a/api-samples/privacy/manifest.json
+++ b/api-samples/privacy/manifest.json
@@ -1,0 +1,10 @@
+{
+  "name": "Privacy API sample",
+  "version": "1.0",
+  "manifest_version": 3,
+  "background": {
+    "service_worker": "service-worker.js"
+  },
+  "permissions": ["privacy"],
+  "action": {}
+}
--- a/api-samples/privacy/service-worker.js
+++ b/api-samples/privacy/service-worker.js
@@ -1,0 +1,27 @@
+// Copyright 2023 Google LLC
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     https://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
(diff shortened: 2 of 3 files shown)


api-samples/privacy/manifest.json
```diff
@@ -0,0 +1,10 @@
+{
+  "name": "Privacy API sample",
+  "version": "1.0",
+  "manifest_version": 3,
+  "background": {
+    "service_worker": "service-worker.js"
+  },
+  "permissions": ["privacy"],
+  "action": {}
```
sebastianbenz, line 9: Is this needed?

sebastianbenz (approved):
Looks good!

sebastianbenz (changes requested):
Sorry I had to take the approval back. Just tried the sample and the behavior is not intuitive. It'd be better to:

1. Show current status on load
2. Toggle the behavior with each action click

api-samples/privacy/manifest.json
sebastianbenz: Ah got, we're programmatically changing the action which is why we don't need more here.

api-samples/privacy/service-worker.js
```diff
@@ -0,0 +1,50 @@
...
+  getAutofillEnabledStatus();
+});
+
+chrome.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
+  if (changeInfo.status === 'complete' && tab.active) {
+    getAutofillEnabledStatus();
+  }
+});
+
+function getAutofillEnabledStatus() {
```
sebastianbenz, line 30: nit: the method is called `getAuto...` but it actually updates the action. Would be good to relect this in the method name.

api-samples/privacy/service-worker.js
```diff
@@ -0,0 +1,50 @@
...
+    const badgeText = autofillEnabled ? 'Enabled' : 'Disabled';
+    const badgeColor = autofillEnabled ? '#00FF00' : '#FF0000';
+
+    chrome.action.setBadgeText({ text: badgeText });
+    chrome.action.setBadgeBackgroundColor({ color: badgeColor });
+  });
+}
+
+chrome.action.onClicked.addListener(() => {
+  chrome.privacy.services.autofillEnabled.get({}, (details) => {
```
sebastianbenz, line 42: Can't we call `getAutofillEnabledStatus` here?

PR #963, opened 2023-06-15
Add scripts for generating sample list

daidr:


.repo/sample-list-generator/README.md
```diff
@@ -0,0 +1,147 @@
+# Sample List Generator
+
+## Overview
+
+It's a script that generates `./extension-samples.json` with the list of all the samples available.
```
oliverdunk, line 5: It would be great to provide some context about why this exists and what it's used for.

.repo/sample-list-generator/src/api-detector.ts
```diff
@@ -0,0 +1,147 @@
+import { IApiItem, TApiTypeResult, TExtensionApiMap } from './types';
+import * as babel from '@babel/core';
+import fs from 'fs/promises';
+import { getAllFiles, singularize } from './utils';
+
+let EXTENSION_API_MAP: TExtensionApiMap = {};
+(async () => {
```
oliverdunk, line 7: Could we move this to a `loadExtensionApis` function maybe that gets called before anything else?

.repo/sample-list-generator/src/api-detector.ts
```diff
@@ -0,0 +1,147 @@
...
+  try {
+    let _EXTENSION_API_MAP = await fs.readFile('./extension-apis.json', 'utf8');
+    EXTENSION_API_MAP = JSON.parse(_EXTENSION_API_MAP);
+  } catch (error) {}
```
oliverdunk, line 11: What's the thinking behind the try/catch? Throwing an error and stopping the script is probably better.

oliverdunk (commented):
This is so cool! I left some initial comments but this is coming along really nicely, great work 👏

.repo/sample-list-generator/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "sample-list-generator",
+  "version": "1.0.0",
+  "scripts": {
+    "start": "ts-node src/index.ts",
+    "prefetch": "ts-node src/fetch.ts"
+  },
+  "devDependencies": {
+    "@types/babel__core": "^7.20.1",
```
sebastianbenz, line 9: Please pin versions to ensure everyone uses exactly the same version.

.repo/sample-list-generator/src/api-detector.ts
```diff
@@ -0,0 +1,144 @@
+import { IApiItem, TApiTypeResult, TExtensionApiMap } from './types';
```
sebastianbenz, line 1: Question: do we need the type prefixes? I prefer to omit these to improve readability (and these are usually provided by the IDE automatically anyway).

.repo/sample-list-generator/src/api-detector.ts
```diff
@@ -0,0 +1,144 @@
...
+): Promise<IApiItem[]> => {
+  // get all js files in the folder
+  const jsFiles = (await getAllFiles(folderPath)).filter((file) =>
+    file.endsWith('.js')
+  );
+
+  const apis: Record<string, Set<string>> = {};
+
+  for (let file of jsFiles) {
+    const apiCalls = await extractApiCalls(await fs.readFile(file));
```
sebastianbenz, line 22: Can we not block on every file to enable parallel execution?

.repo/sample-list-generator/src/api-detector.ts
```diff
@@ -0,0 +1,144 @@
...
+): Promise<Record<string, string[]>> => {
+  return new Promise((resolve) => {
+    const calls: Record<string, string[]> = {};
+
+    babel.parse(
+      file.toString('utf8'),
+      { ast: true, compact: false },
+      (err, result) => {
+        if (err || !result) {
+          resolve(calls);
```
sebastianbenz, line 59: Why not reject if there's an error?

.repo/sample-list-generator/src/api-detector.ts
```diff
@@ -0,0 +1,144 @@
...
+          return;
+        }
+
+        babel.traverse(result, {
```
sebastianbenz, line 63: This is quite a long method, can you please split it up to make it easier to understand.

.repo/sample-list-generator/src/utils.ts
```diff
@@ -0,0 +1,59 @@
+import fs from 'fs/promises';
```
sebastianbenz, line 1: Can you please split this file up into more meaningful parts, util classes are an antipattern as they quickly grow and break separation of concerns.

.repo/sample-list-generator/src/index.ts
```diff
@@ -0,0 +1,75 @@
+import path from 'path';
+import fs from 'fs/promises';
+
+import { AVAILABLE_FOLDERS, REPO_BASE_URL } from './constants';
+import { getApiListForSample } from './api-detector';
+import type { IAvailableFolderItem, ISampleItem } from './types';
+import { getBasePath, isFileExists } from './utils';
+
+const getAllSamples = async () => {
```
sebastianbenz, line 9: Please move sample collection into a separate class, index.ts should only orchestrate and not include any actual logic.

.repo/sample-list-generator/src/fetch.ts
```diff
@@ -0,0 +1,70 @@
+import fetch from 'node-fetch';
```
sebastianbenz, line 1: The file name fetch is a bit misleading as the code seems to do something completely different. Can you find a. better name?

.repo/sample-list-generator/src/fetch.ts
```diff
@@ -0,0 +1,70 @@
...
+import path from 'path';
+import fs from 'fs/promises';
+import { JSDOM } from 'jsdom';
+import { TExtensionApiMap } from './types';
+
+const getApiNames = async () => {
+  const re =
+    /<tr><td><a class="link weight-medium" href=(\S+)>(\S+)<\/a><\/td><td class="reference-collapse type">/g;
+  const rawHtml = await (
+    await fetch('https://developer.chrome.com/docs/extensions/reference/')
```
sebastianbenz, line 11: It'd be good to check for the error code here, otherwise we might parse a 503 error page...

.repo/sample-list-generator/extension-apis.json
```diff
@@ -0,0 +1,1994 @@
+{
```
sebastianbenz, line 1: Can you add a flag to this file indicating that this is auto generated and should not be edited?

sebastianbenz (changes requested):
Left a few drive-by comments. High level feedback: this is a great start, but it'd be good to make sure that this code is easy to maintain in the future. I've noticed a few things that could be improved:

- properly naming things
- avoiding large methods / classes
- splitting things up based on their responsibilities
- add tests: please make sure that the critical parts have tests, in particular the part that parses samples for API usage etc would greatly benefit from tests.

Thanks!

.repo/sample-list-generator/package.json
```diff
@@ -0,0 +1,20 @@
+{
+  "name": "sample-list-generator",
```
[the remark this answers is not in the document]
daidr: I think the relation between this script and the reo is not significant. If some dependencies are placed in the package.json at the root, it may cause some misunderstandings.

.repo/sample-list-generator/extension-apis.json
oliverdunk: This sounds good but I don't think we should do a comment since only certain JSON parsers support that. Are you thinking something like this?

```json
{
  "_comment": "This file is autogenerated. Do not edit.",
  ...
}
```

.repo/sample-list-generator/src/api-detector.ts
oliverdunk: Agreed on removing these, it's not a pattern I've found particularly useful in the past :)

.repo/sample-list-generator/src/api-detector.ts
oliverdunk: I think we can remove the async function call as well and just have `loadExtensionApis()` in the top/bottom of the JS file. Then that function can update `EXTENSION_API_MAP`. Let me know if it would help for me to explain that more :)

oliverdunk (changes requested):
Left a few comments and I see Sebastian has some too. I think those are a good start and then I can add some more feedback :)

.repo/sample-list-generator/src/constants.ts
```diff
@@ -0,0 +1,16 @@
+export type AvailableFolderTypes = (typeof AVAILABLE_FOLDERS)[number]['type'];
```
oliverdunk, line 1: Thoughts on just defining this as `"API_SAMPLE" | "FUNCTIONAL_SAMPLE"`? I think that's better, because otherwise a typo in `AVAILABLE_FOLDERS` will cause a new type to suddenly appear.

I would also suggest renaming this to just `FolderType`.

.repo/sample-list-generator/src/constants.ts
```diff
@@ -0,0 +1,16 @@
...
+export const AVAILABLE_FOLDERS = [
+  {
+    path: 'api-samples',
+    type: 'API_SAMPLE'
+  },
+  {
+    path: 'functional-samples',
+    type: 'FUNCTIONAL_SAMPLE'
+  }
+] as const;
```
oliverdunk, line 13: Once we resolve the above comment we could type this as `{ path: string, type: FolderType }[]` and we should get autocomplete in some IDEs :)

.repo/sample-list-generator/src/index.ts
```diff
@@ -0,0 +1,15 @@
+import path from 'path';
+import fs from 'fs/promises';
+import { getAllSamples } from './libs/sample-collector';
+
+(async () => {
```
oliverdunk, line 5: Similar to the other file can we remove this and just have a `start()` function or similar? I'm not a huge fan of these immediately-invoked functions.

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,174 @@
...
+import { ParallelController } from '../utils/parallel';
+
+const CPU_CORES = os.cpus().length;
+
+let EXTENSION_API_MAP: ExtensionApiMap = loadExtensionApis();
+
+export const getApiType = (
+  apiCategory: string,
+  apiName: string
+): ApiTypeResult => {
```
oliverdunk, line 17: Would `api` or `namespace` and `propertyName` be better names? If so we could make that change across the code.

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,174 @@
...
+  apiCategory = apiCategory.replace(/_/g, '.');
+
+  if (EXTENSION_API_MAP[apiCategory]) {
+    const apiTypes = EXTENSION_API_MAP[apiCategory];
+
+    for (let type of Object.keys(apiTypes)) {
```
oliverdunk, line 23: I actually think individually checking things here might be easier to read. e.g:

```ts
if (apiTypes.functions.includes(apiName) {
  return "method";
}

if (apiTypes.properties.includes(apiName) {
  return "property";
}

...
```

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,174 @@
...
+      if (apiTypes[type].includes(apiName)) {
+        return type as ApiTypeResult;
+      }
+    }
+  }
+  console.log('api not found', apiCategory, apiName);
+  return 'unknown';
+};
+
+const getNextLevelProperty = (
```
oliverdunk, line 33: Hmm, I'm not sure what this does. Could you make that clearer in the code with some comments and maybe renaming the function?

.repo/sample-list-generator/src/utils/parallel.ts
```diff
@@ -0,0 +1,29 @@
+export class ParallelController {
```
oliverdunk, line 1: Would you mind sharing your thoughts on adding this? It's super cool, but feels like maybe a bit much. I feel like we should be fine just calling `Promise.all` and trusting the system to figure out how to run things.

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,174 @@
...
+  });
+
+  await Promise.all(parallelHandler);
+
+  const result: ApiItem[] = [];
+
+  Object.keys(apis).forEach((apiType) => {
+    apis[apiType].forEach((apiName) => {
+      result.push({
+        type: singularize(apiType) as ApiTypeResult,
```
oliverdunk, line 78: Why do we need `singularize` here?

.repo/sample-list-generator/src/libs/api-loader.ts
```diff
@@ -0,0 +1,20 @@
...
+  if (!isFileExistsSync(filePath)) {
+    console.error(
+      'extension-apis.json does not exist. Please run "npm run prepare-chrome-types" first.'
+    );
+    process.exit(1);
+  }
+
+  let _EXTENSION_API_MAP = fs.readFileSync(filePath, 'utf8');
+  let EXTENSION_API_MAP = JSON.parse(_EXTENSION_API_MAP) as ExtensionApiMap;
+  return EXTENSION_API_MAP;
```
oliverdunk, line 19: Something like this probably works right? We don't need the `as` as far as I know because JSON.parse returns any, so it should be fine to just guess that from the function return signature.

```suggestion
  let data = fs.readFileSync(filePath, 'utf8');
  return JSON.parse(data);
```

.repo/sample-list-generator/src/utils/manifest.ts
```diff
@@ -0,0 +1,21 @@
+import fs from 'fs/promises';
+import { ManifestMetadata } from '../types';
+
+export const getManifestMetadata = async (
+  manifesPath: string,
+  defaultName: string
+): Promise<ManifestMetadata> => {
+  const manifest = await fs.readFile(manifesPath, 'utf8');
+  const parsedManifest = JSON.parse(manifest);
+  const {
```
oliverdunk, line 10: Do we need this line? I feel like we could just do it all in the return.

.repo/sample-list-generator/src/utils/manifest.ts
```diff
@@ -0,0 +1,21 @@
...
+    name: title = defaultName,
+    description = '',
+    permissions = []
```
oliverdunk, line 13: Are we planning to load permissions here?

.repo/sample-list-generator/test/api-detector/api-detector.test.ts
```diff
@@ -0,0 +1,100 @@
+import { describe, it, beforeEach } from 'mocha';
```
oliverdunk, line 1: Love these tests, great work!

.repo/sample-list-generator/src/types.ts
```diff
@@ -0,0 +1,36 @@
...
+}
+
+export interface ManifestMetadata {
+  title: string;
+  description: string;
+  permissions: string[];
+}
+
+export type SampleItem = {
+  type: 'API_SAMPLE' | 'FUNCTIONAL_SAMPLE';
```
oliverdunk, line 16: This can probably use AvailableFolderTypes right?

.repo/sample-list-generator/src/types.ts
```diff
@@ -0,0 +1,36 @@
...
+} & ManifestMetadata;
+
+export interface AvailableFolderItem {
+  path: string;
+  type: AvailableFolderTypes;
+}
+
+export type ApiTypeResult =
+  | 'event'
+  | 'method'
```
oliverdunk, line 29: Can we use `function`? I know it doesn't have that name elsewhere, but it's more accurate for JS.

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,174 @@
...
+      { ast: true, compact: false },
+      (err, result) => {
+        if (err || !result) {
+          reject(err);
+          return;
+        }
+
+        babel.traverse(result, {
+          MemberExpression(path) {
+            const object = path.node.object;
```
oliverdunk, line 105: As @sebastianbenz mentioned it'd be nice to try and find a way to break this down more. Let me know if you need help thinking that through.

oliverdunk (changes requested):
Looking great still, love the tests! Left a few more comments.

.repo/sample-list-generator/src/libs/api-detector.ts
daidr: Because the getApiType function returns the raw key, such as "methods". I would like to change it to "method", perhaps I should do this conversion in the getApiType function?

.repo/sample-list-generator/src/types.ts
daidr: I noticed that developer.chrome.com uses the term "method(s)". Perhaps using "method(s)" would be easier to understand.🤔

oliverdunk: I think you're right actually. I was thinking that we should slowly change developer.chrome.com to say functions too, but looking in to it, it does seem like even MDN uses the term "method" for a function on an object property.

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,156 @@
...
+
+            // check if expression isn't browser.xxx or chrome.xxx
+            if (
+              object.type !== 'Identifier' ||
+              !['browser', 'chrome'].includes(object.name)
+            ) {
+              return;
+            }
+
+            const parts = getFullMemberExpression(path);
```
oliverdunk, line 139: Can we update `getFullMemberExpression` to include "chrome" or "browser", and then use it to help with the line above?

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,156 @@
...
+      )
+    );
+  }
+
+  return result;
+}
+
+export function getApiItem(parts: string[]): ApiItem {
+  // special case for `chrome.storage`
+  if (parts[0] === 'storage') {
```
oliverdunk, line 85: If we get rid of this, what happens? I'm assuming it means we will no longer get "storage.onChanged" for "storage.local.onChanged" but I'm not sure how important that is and I feel like getting rid of the special case would be nice.

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,156 @@
...
+
+            // api not found
+            if (apiType === 'unknown') {
+              return;
+            }
+            calls.push({ type: apiType, namespace, propertyName });
+          }
+        });
+
+        resolve(uniqueItems(calls));
```
oliverdunk, line 152: Can we avoid `uniqueItems` by building the map of what we've seen in this function?

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,156 @@
...
+            }
+
+            const parts = getFullMemberExpression(path);
+
+            const { namespace, propertyName } = getApiItem(parts);
+            let apiType = getApiType(namespace, propertyName);
+
+            // api not found
+            if (apiType === 'unknown') {
+              return;
```
oliverdunk, line 146: If we add a console.warn here with parts.join(".") does that create a lot of spam or not really?

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,156 @@
...
+              object.type !== 'Identifier' ||
+              !['browser', 'chrome'].includes(object.name)
+            ) {
+              return;
+            }
+
+            const parts = getFullMemberExpression(path);
+
+            const { namespace, propertyName } = getApiItem(parts);
+            let apiType = getApiType(namespace, propertyName);
```
oliverdunk, line 142: Nit: Rename `apiType` to just `type`?

.repo/sample-list-generator/src/utils/manifest.ts
```diff
@@ -0,0 +1,21 @@
+import fs from 'fs/promises';
+import { ManifestMetadata } from '../types';
+
+export const getManifestMetadata = async (
+  manifesPath: string,
```
oliverdunk, line 5: Nit: Typo, should be manifestPath :)

.repo/sample-list-generator/src/libs/sample-collector.ts
```diff
@@ -0,0 +1,69 @@
...
+  let samples: SampleItem[] = [];
+
+  // loop through all available folders
+  // e.g. api-samples, functional-samples
+  for (let samplesFolder of AVAILABLE_FOLDERS) {
+    const currentSamples = await getSamples(
+      samplesFolder.path,
+      samplesFolder.type
+    );
+    samples = [...samples, ...currentSamples];
```
oliverdunk, line 19: Can we just do samples.push(...samples)?

.repo/sample-list-generator/src/prepare-chrome-types.ts
```diff
@@ -0,0 +1,68 @@
...
+    const apiDetails: ExtensionApiMap[string] = {
+      properties: [],
+      methods: [],
+      types: [],
+      events: []
+    };
+
+    for (let variable of chromeApiDetails._type.properties) {
+      const name = variable.name as string;
+      const type = getTypeFromPageId(variable._pageId) as string;
```
oliverdunk, line 45: Can we just use `variable.kind` here? I think the mapping is here: https://unpkg.com/typedoc@0.24.8/dist/lib/models/reflections/kind.d.ts

.repo/sample-list-generator/src/prepare-chrome-types.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  for (const [chromeApiKey, chromeApiDetails] of Object.entries(chromeTypes)) {
+    const apiDetails: ExtensionApiMap[string] = {
+      properties: [],
+      methods: [],
+      types: [],
+      events: []
+    };
+
+    for (let variable of chromeApiDetails._type.properties) {
```
oliverdunk, line 43: I think renaming `variable` to `property` might be nicer?

oliverdunk (changes requested):
Left a bunch of extra comments - some great improvements from last time though, I think the Babel changes really tidied things up.

.repo/sample-list-generator/src/libs/api-detector.ts
daidr: If this part of the logic is moved to getFullMemberExpression, it may cause unnecessary recursion. For example, for chrome.storage.onChanged.addListener, the former will only recursively generate ["storage", "onChanged", "addListener"], but if the above logic is removed, multiple arrays will be recursively generated: ["chrome", "storage", "onChanged", "addListener"], ["onChanged", "addListener"], ["addListener"].

.repo/sample-list-generator/src/prepare-chrome-types.ts
daidr: I think `variable.kind` may not be sufficient to determine the type of API in the doc. 🤔
I ran a script on the existing APIs, and the corresponding types and `variable.kind` are as follows: 

| Type | Kinds |
|--------|--------|
| property | 32 |
| method | 64 |
| type | 256/4194304/128 | 
| event | 32 |

oliverdunk: It looks like in https://github.com/GoogleChrome/developer.chrome.com/blob/215f8ef57819f451ecf6d5fdd84b9981f60a0a10/external/build/lib/dts-parse.js#L33 we identify events by checking `variable.name` against a list of special types. Would that work?

daidr: It works! However, I wonder if it's worth implementing the logic again in the script. Will it make understanding or maintenance difficult?

.repo/sample-list-generator/src/libs/api-detector.ts
daidr: It seems that there is indeed no such usage method in the document. I will remove this special case.

daidr: Oh, I've thought about it and maybe this special case shouldn't be removed: 
`storage.sync` should correspond to `/reference/storage/#property-sync`, but `storage.sync.onChange` should correspond to `/reference/storage/#event-onChanged`, which is different from other APIs. However, my code can handle the situation of `storage.onChange`.

.repo/sample-list-generator/src/libs/api-detector.ts
daidr: I think `uniqueItems` here can be removed directly. It only retrieves the APIs called by individual files, and after merging, another deduplication process is still needed.

.repo/sample-list-generator/src/prepare-chrome-types.ts
oliverdunk: I think it's better than looking at the page URLs 😄

daidr: > I think it's better than looking at the page URLs 😄

Got it, I will try to use `variable.kind` and expect to finish it today.

.repo/sample-list-generator/src/libs/api-detector.ts
oliverdunk: Hmm. Shouldn't it actually link to https://developer.chrome.com/docs/extensions/reference/storage/#event-StorageArea-onChanged?

.repo/sample-list-generator/README.md
```diff
@@ -0,0 +1,147 @@
...
+  permissions: string[];
+}
+
+// the type of extension-samples.json file is ISampleItem[]
+```
+
+## Example
+
+```json
+// extension-samples.json
```
oliverdunk, line 58: Can we move this out of the code block? It doesn't seem to render well.

.repo/sample-list-generator/README.md
```diff
@@ -0,0 +1,147 @@
...
+  type: 'API_SAMPLE' | 'FUNCTIONAL_SAMPLE';
+  name: string;
+  title: string;
+  description: string;
+  repo_link: string;
+  apis: ApiItem[];
+  permissions: string[];
+}
+
+// the type of extension-samples.json file is ISampleItem[]
```
oliverdunk, line 52: Could we update `ISampleItem` here (and check the types above as well)?

.repo/sample-list-generator/src/index.ts
```diff
@@ -0,0 +1,17 @@
...
+const start = async () => {
+  const samples = await getAllSamples();
+
+  // write to extension-samples.json
+  await fs.writeFile(
+    path.join(__dirname, '../extension-samples.json'),
+    JSON.stringify(samples, null, 2)
+  );
+
+  console.log('Done!');
```
oliverdunk, line 14: I think we can probably drop this?

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,161 @@
...
+  sampleFolderPath: string
+): Promise<ApiItemWithType[]> => {
+  // get all js files in the folder
+  const jsFiles = (await getAllFiles(sampleFolderPath)).filter((file) =>
+    file.endsWith('.js')
+  );
+
+  const calls: ApiItemWithType[] = [];
+
+  const parallelHandler = jsFiles.map(async (file) => {
```
oliverdunk, line 50:
```suggestion
  await Promise.all(jsFiles.map(async (file) => {
```

I think something like this probably reads well enough and avoids creating a temporary variable?

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,161 @@
...
+    const _calls = await extractApiCalls(await fs.readFile(file));
```
oliverdunk, line 51: Maybe we could call this `callsFromFile` or similar?

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,161 @@
...
+    return {
+      namespace: 'storage',
+      propertyName: parts.includes('onChanged') ? 'onChanged' : parts[1]
+    };
+  }
+
+  let namespace = '';
+  let propertyName = '';
+
+  if (EXTENSION_API_MAP[`${parts[0]}.${parts[1]}`]) {
```
oliverdunk, line 104: Can we add a comment about why this is needed?

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,161 @@
...
+  }
+
+  return { namespace, propertyName };
+}
+
+function uniqueItems(array: ApiItemWithType[]) {
+  const tmp = new Map();
+  return array.filter((item) => {
+    const fullApiString = `${item.namespace}.${item.propertyName}`;
+    return !tmp.has(fullApiString) && tmp.set(fullApiString, 1);
```
oliverdunk, line 119: Could this be a `Set` rather than a `Map`, since we only have one value?

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,161 @@
...
+  const parallelHandler = jsFiles.map(async (file) => {
+    const _calls = await extractApiCalls(await fs.readFile(file));
+    calls.push(..._calls);
+  });
+  await Promise.all(parallelHandler);
+
+  return uniqueItems(calls);
+};
+
+export function getFullMemberExpression(
```
oliverdunk, line 59: Could you go through and try to add a comment to each function in this PR? It'd be super helpful so new people know what everything is doing I think :)

oliverdunk (changes requested):
Left a few more comments but I think this is getting really close. Hopeful we might be able to merge this week 🤞

.repo/sample-list-generator/src/libs/api-detector.ts
daidr: Oh sorry, this is my mistake. I overlooked this. 😥

.repo/sample-list-generator/package.json
```diff
@@ -0,0 +1,24 @@
...
+  "version": "1.0.0",
+  "scripts": {
+    "start": "ts-node src/index.ts",
+    "prepare-chrome-types": "ts-node src/prepare-chrome-types.ts",
+    "test": "mocha --require ts-node/register test/**/*.test.ts"
+  },
+  "devDependencies": {
+    "@types/babel__core": "7.20.1",
+    "@types/mocha": "^10.0.1",
```
sebastianbenz, line 11: Please pin dependencies

.repo/sample-list-generator/src/constants.ts
```diff
@@ -0,0 +1,16 @@
+export type FolderTypes = "API_SAMPLE" | "FUNCTIONAL_SAMPLE";
+
+// Define all available folders for samples
+export const AVAILABLE_FOLDERS = [
+  {
+    path: 'api-samples',
+    type: 'API_SAMPLE'
```
sebastianbenz, line 7: nit: Why not use the type defined above?

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,201 @@
...
+ * getApiType('tabs', 'query')
+ * // returns 'method'
+ */
+export const getApiType = (
+  namespace: string,
+  propertyName: string
+): ApiTypeResult => {
+  namespace = namespace.replace(/_/g, '.');
+
+  if (EXTENSION_API_MAP[namespace]) {
```
sebastianbenz, line 30: nit: extract `EXTENSION_API_MAP[namespace]` to a variable

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,201 @@
...
+/**
+ * Gets all the api calls in a sample.
+ * @param sampleFolderPath - The path to the sample folder.
+ * @returns A promise that resolves to an array of apis the sample uses.
+ */
+export const getApiListForSample = async (
+  sampleFolderPath: string
+): Promise<ApiItemWithType[]> => {
+  // get all js files in the folder
+  const jsFiles = (await getAllFiles(sampleFolderPath)).filter((file) =>
```
sebastianbenz, line 58: You could move this into a method in the filesystem util.

.repo/sample-list-generator/src/libs/api-detector.ts
```diff
@@ -0,0 +1,201 @@
...
+  const tmp = new Set<string>();
+  return array.filter((item) => {
+    const fullApiString = `${item.namespace}.${item.propertyName}`;
+    return !tmp.has(fullApiString) && tmp.add(fullApiString);
+  });
+}
+
+/**
+ * Extracts all chrome and browser api calls from a file.
+ * @param file - The file to extract api calls from.
```
sebastianbenz, line 157: nit: why is this not named buffer? 

I think it'd be better to change the method to accept a string instead of a buffer this moves responsibility to the caller.

.repo/sample-list-generator/src/libs/sample-collector.ts
```diff
@@ -0,0 +1,72 @@
...
+  const basePath = getBasePath();
+
+  // get all contents in the folder
+  const contents = await fs.readdir(path.join(basePath, currentRootFolderPath));
+
+  for (let content of contents) {
+    const currentPath = path.join(basePath, currentRootFolderPath, content);
+
+    // if content is not a folder, skip
+    if (!(await fs.stat(currentPath)).isDirectory()) {
```
sebastianbenz, line 39: maybe turn this into a helper in filesystem?

.repo/sample-list-generator/README.md
```diff
@@ -0,0 +1,147 @@
...
+
+```bash
+npm run prepare-chrome-types
+```
+
+### Run the generator
+
+```bash
+npm start
+```
```
sebastianbenz, line 29: Explain how to run tests

sebastianbenz (approved):
Looks great - thanks a lot! Left a few nits, but nothing blocking.

.repo/sample-list-generator/src/constants.ts
daidr: The assertion here is unnecessary. Could I change it to the following code?

```ts
export const AVAILABLE_FOLDERS: { path: string, type: FolderTypes }[] =
```

oliverdunk: Removing the assertion seems good. I don't think it's possible to use the type directly because it's just a string union.

.repo/sample-list-generator/README.md
daidr: Done. https://github.com/GoogleChrome/chrome-extensions-samples/pull/963/commits/b316e888c2e88c6fab2c51b1a102138b4c49d6ee

.repo/sample-list-generator/src/libs/api-detector.ts
daidr: Done. https://github.com/GoogleChrome/chrome-extensions-samples/pull/963/commits/709b196051f32d94da42fc23ff21c7a96a8eccac

.repo/sample-list-generator/package.json
daidr: Done. https://github.com/GoogleChrome/chrome-extensions-samples/pull/963/commits/4a88546743ffdc78a352c2f0f30075e4312a7d51

.repo/sample-list-generator/src/libs/api-detector.ts
daidr: Done. https://github.com/GoogleChrome/chrome-extensions-samples/pull/963/commits/f937230468614ee77c74d57d6174e7460fe2941b

.repo/sample-list-generator/src/libs/api-detector.ts
daidr: Done. https://github.com/GoogleChrome/chrome-extensions-samples/pull/963/commits/4fcffeefe325e859cf97af19f0b8af87c3cdff5f & https://github.com/GoogleChrome/chrome-extensions-samples/pull/963/commits/0c663a5dbfb4efc4b48de28dd822a6a9f1aea504

.repo/sample-list-generator/src/libs/sample-collector.ts
daidr: Done. https://github.com/GoogleChrome/chrome-extensions-samples/pull/963/commits/2bd89a80506df4747af88a394f9f7e15f212bf76

.repo/sample-list-generator/src/constants.ts
daidr: Done. https://github.com/GoogleChrome/chrome-extensions-samples/pull/963/commits/bd56b470d789136a3260124b355e19ef22ae586c

oliverdunk (approved):
LGTM! Thanks so much for all of the hard work on this :)

PR #983, opened 2023-07-12
[SidePanel API] Add sidePanel.open() cookbook

AmySteam:
Adding a sample for the sidePanel API.

Related https://github.com/GoogleChromeLabs/Extension-Docs/issues/143


AmySteam:
@oliverdunk Can you wait until tomorrow or Friday to review this? I tried other experiments and would like to switch it up a bit.

AmySteam:
@oliverdunk consider yourself poked 👉. I explain how it works in the readme. Let me know what you think about the use cases. I was trying to stay away from action click and shortcut since we already have a sample for that. But I'm open to feedback 🥸🦖

functional-samples/cookbook.sidepanel-open/README.md
```diff
@@ -0,0 +1,25 @@
+# Opening the side panel through a user interaction
```
AmySteam, line 1: @sebastianbenz Do you think I'm demonstrating too many use cases in one sample? Do you think I should narrow it down to one?

sebastianbenz: Nope, I think this is perfect

functional-samples/cookbook.sidepanel-open/service-worker.js
```diff
@@ -0,0 +1,44 @@
...
+chrome.contextMenus.onClicked.addListener((info, tab) => {
+  if (info.menuItemId === 'openSidePanel') {
+    // This will open the panel in all the pages on the current window.
+    chrome.sidePanel.open({ windowId: tab.windowId });
+  }
+});
+
+chrome.runtime.onMessage.addListener((message, sender) => {
+  // The callback for runtime.onMessage must return falsy if we're not sending a response
+  (async () => {
```
sebastianbenz, line 33: Curious: why do we need the nested async function here?

sebastianbenz (approved):
Looks good!

functional-samples/cookbook.sidepanel-open/service-worker.js
AmySteam: We can send a message response using the third callback argument, a sendResponse function.

Not every callback will send a response, and therefore we need to tell Chrome if we're going to send a response. Chrome uses the return value of the callback to determine whether to expect a response or not: if the return value is truthy, we have to call the sendResponse function.

This means we can't use an async function as our callback function because async functions return a Promise object (which is a truthy value, regardless of what value the Promise will contain). 

Therefore, if we're going to do something async but not send a message response, we need to find a way to return something like undefined.

Admin310:
yo*

oliverdunk:
Sorry for the late review, but this looks great! Awesome work 👏

PR #985, opened 2023-07-15
Fix rename file in debugger example app

youngkyo0504:
there is incorrect value in `manifest.json`


patrickkettner (approved):

patrickkettner:
thank you!

PR #986, opened 2023-07-18
Add new sample: no-cookies rule manager

daidr:
Added a new sample to demonstrate how to dynamically add and remove DNR rules.


api-samples/declarativeNetRequest/no-cookies-manager/manifest.json
```diff
@@ -0,0 +1,23 @@
...
+  "action": {
+    "default_title": "no-cookies Rule Manager",
+    "default_popup": "popup.html"
+  },
+  "host_permissions": ["<all_urls>"],
+  "permissions": [
+    "declarativeNetRequest",
+    "declarativeNetRequestFeedback",
+    "activeTab"
+  ],
```
AmySteam, line 18: @oliverdunk can you double-check that all these permissions are needed?

daidr: I checked again and found that the `activeTab` permission is not necessary. I will correct it right away.

api-samples/declarativeNetRequest/no-cookies-manager/service_worker.js
```diff
@@ -0,0 +1,45 @@
+chrome.runtime.onInstalled.addListener(async function () {
+  // restore the default rule if the extension is installed or updated
+  const _rules = await chrome.declarativeNetRequest.getDynamicRules();
```
oliverdunk, line 3: I think we can just call this rules or existingRules rather than needing the underscore?

api-samples/declarativeNetRequest/no-cookies-manager/popup.js
```diff
@@ -0,0 +1,24 @@
...
+const matchedRulesList = document.getElementById('matchedRulesList');
+
+async function init() {
+  const tabId = (await chrome.tabs.query({ active: true }))[0].id;
+  const result = await chrome.declarativeNetRequest.getMatchedRules({ tabId });
+  if (result.rulesMatchedInfo.length === 0) {
+    document.querySelector('.tips').style.display = 'block';
+    return;
+  }
+  document.querySelector('.matched-rules').style.display = 'block';
```
oliverdunk, line 14: We should probably have some logic to hide the tips if everything works?

api-samples/declarativeNetRequest/no-cookies-manager/manager.js
```diff
@@ -0,0 +1,190 @@
...
+});
+
+function appendRuleItem(
+  container,
+  { id, conditionType, conditionValue, caseSensitive }
+) {
+  // Create a new rule item from the template
+  let ruleItem = ruleItemTemplate.content.cloneNode(true);
+  container.appendChild(ruleItem);
+  ruleItem = container.lastElementChild;
```
oliverdunk, line 33: Is this needed?

api-samples/declarativeNetRequest/no-cookies-manager/manager.js
```diff
@@ -0,0 +1,190 @@
...
+    // For the regex filter, verify if the regex is supported
+    if (_conditionType === 'regexFilter') {
+      const result = await chrome.declarativeNetRequest.isRegexSupported({
+        isCaseSensitive: _caseSensitive,
+        regex: _conditionValue
+      });
+      if (!result.isSupported) {
+        // If the regex is invalid, disable the save button
+        setSaveButtonDisabled();
+        alert(`Invalid regex: ${result.reason}`);
```
oliverdunk, line 129: Isn't this going to alert for every character you type? That seems like it might be a bit aggressive.

api-samples/declarativeNetRequest/no-cookies-manager/manager.js
daidr: I will modify it into the following code:
```js
const ruleItem = ruleItemTemplate.content.cloneNode(true).children[0];
container.appendChild(ruleItem);
```

Because I need to manipulate DOM elements, not the document fragment itself.

api-samples/declarativeNetRequest/no-cookies-manager/manager.js
daidr: I'm not sure, but I think the `change` event will only be triggered when the input box loses focus. The event that is triggered every time there is an input should be the `input` event.

api-samples/declarativeNetRequest/no-cookies-manager/service_worker.js
daidr: Done. https://github.com/GoogleChrome/chrome-extensions-samples/pull/986/commits/4b52b3dd127657998006b6318925836ba958f248

api-samples/declarativeNetRequest/no-cookies-manager/popup.js
daidr: image
Actually, `matched-rules` is the container that holds the results. I think it might be better to name it `matched-rules-wrapper`?

oliverdunk: Sorry, my comment was probably on an unhelpful line. I was just wondering if we should add this?

```
document.querySelector('.tips').style.display = 'none';
```

Otherwise I think the tips will never go away?

api-samples/declarativeNetRequest/no-cookies-manager/manager.js
oliverdunk: If that works that seems better 👍

api-samples/declarativeNetRequest/no-cookies-manager/manager.js
oliverdunk: You're right! I thought I deleted this comment actually, but I guess I didn't.

api-samples/declarativeNetRequest/no-cookies-manager/manager.html
```diff
@@ -0,0 +1,83 @@
...
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>no-cookies Rule Manager</title>
+    <link rel="stylesheet" href="./manager.css" />
+  </head>
+
+  <body>
+    <div class="toolbar">
+      <button id="addRuleButton">Add Rule</button>
+      <button id="addRuleButton">View Current Rule List</button>
```
oliverdunk, line 13: We probably don't want two buttons with the same ID?

api-samples/declarativeNetRequest/no-cookies-manager/manager.html
```diff
@@ -0,0 +1,83 @@
...
+        carrying cookies.
+      </small>
+    </div>
+    <div id="rulesList">
+      <!-- Rule Items will be added here -->
+    </div>
+
+    <template id="ruleItemTemplate">
+      <div class="rule-item">
+        <div>
```
oliverdunk, line 25: Could we use a label element here?

api-samples/declarativeNetRequest/no-cookies-manager/popup.js
daidr: Oh, I understand what you mean. 
But I think this tips itself should not be removed. When the user opens the popup, if the extension does not find a matching rule, this element should always be present during the time the popup is open.

oliverdunk: Ah, ok. That makes sense!

oliverdunk:
Could we move this to the functional samples? Something like `functional-samples/sample.dnr-rule-manager` probably works well.

Admin310:
👌Sent from my Galaxy
-------- Original message --------From: Xuezhou Dai ***@***.***> Date: 7/28/23  16:00  (GMT+06:30) To: GoogleChrome/chrome-extensions-samples ***@***.***> Cc: Subscribed ***@***.***> Subject: Re: [GoogleChrome/chrome-extensions-samples] Add new sample: no-cookies rule manager (PR #986) 
@daidr commented on this pull request.



In api-samples/declarativeNetRequest/no-cookies-manager/popup.js:
> @@ -0,0 +1,24 @@
+document.getElementById('openManagerTab').addEventListener('click', () => {
+  chrome.tabs.create({ url: 'manager.html' });
+});
+
+const matchedRulesList = document.getElementById('matchedRulesList');
+
+async function init() {
+  const tabId = (await chrome.tabs.query({ active: true }))[0].id;
+  const result = await chrome.declarativeNetRequest.getMatchedRules({ tabId });
+  if (result.rulesMatchedInfo.length === 0) {
+    document.querySelector('.tips').style.display = 'block';
+    return;
+  }
+  document.querySelector('.matched-rules').style.display = 'block';

Oh, I understand what you mean.
But I think this tips itself should not be removed. When the user opens the popup, if the extension does not find a matching rule, this element should always be present during the time the popup is open.

functional-samples/sample.dnr-rule-manager/manager.js
```diff
@@ -0,0 +1,199 @@
...
+      await removeRule(id);
+    }
+    ruleItem.remove();
+  });
+  ruleItem.querySelector('.save-rule').addEventListener('click', async () => {
+    const {
+      conditionValue: _conditionValue,
+      conditionType: _conditionType,
+      caseSensitive: _caseSensitive
+    } = getCurrentRuleValues();
```
oliverdunk, line 79: Do we need to rename these? It seems like we shouldn't have to, which should hopefully clean up the code quite a bit.

functional-samples/sample.dnr-rule-manager/manager.js
```diff
@@ -0,0 +1,199 @@
...
+    });
+    // Update the rule ID
+    ruleItem.querySelector('.rule-id').value = _id;
+    // Disable the save button
+    setSaveButtonDisabled();
+    // Update parameter
+    id = _id;
+    conditionType = _conditionType;
+    conditionValue = _conditionValue;
+    caseSensitive = _caseSensitive;
```
oliverdunk, line 95: Do these lines do anything?

daidr: Perhaps I can rename these variables starting with an underscore, for example changing `_conditionValue` to `currentConditionValue`.

functional-samples/sample.dnr-rule-manager/README.md
```diff
@@ -0,0 +1,28 @@
+# chrome.declarativeNetRequest - No Cookies Rule Manager
+
+This sample demonstrates using the [`chrome.declarativeNetRequest`](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/) API to remove the "Cookie" header from requests matching custom url filters.
```
jpmedley, line 3:
```suggestion
This sample demonstrates using the [`chrome.declarativeNetRequest`](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/) API to remove the `Cookie` header from requests matching custom URL filters.
```

functional-samples/sample.dnr-rule-manager/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Overview
+
+Once this extension is installed, there will be two default rules:
```
jpmedley, line 7:
```suggestion
This extension has two default rules:
```

functional-samples/sample.dnr-rule-manager/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+1. Any main frame or XHR requests ending with `?nocookies=1` will have their cookies removed (URL Filter).
+
+2. Any main frame or XHR requests matching `.*\.google\.com` will have their cookies removed (Regex Filter).
```
jpmedley, line 11:
```suggestion
* Any main frame or XHR requests ending with `?nocookies=1` will have their cookies removed (URL Filter).

* Any main frame or XHR requests matching `.*\.google\.com` will have their cookies removed (Regex Filter).
```

jpmedley: Corporate styleguide reserves numbers for procedures.

functional-samples/sample.dnr-rule-manager/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+For example, install this extension and try navigating to <https://github.com/GoogleChrome/chrome-extensions-samples?no-cookies=1> or <https://www.google.com> You should appear signed out. The number of requests modified by this extension will be displayed on the extension's badge.
+
+You can edit these rules in the manager page.
+
+## Implementation Notes
+
+[`chrome.declarativeNetRequest.setExtensionActionOptions`](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#method-setExtensionActionOptions) is used to set the extension's badge text to the number of requests modified by the extension.
```
jpmedley, line 19:
```suggestion
[`chrome.declarativeNetRequest.setExtensionActionOptions()`](https://developer.chrome.com/docs/extensions/reference/declarativeNetRequest/#method-setExtensionActionOptions) sets the extension's badge text to the number of requests modified by the extension.
```

functional-samples/sample.dnr-rule-manager/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+`declarativeNetRequestFeedback` permission is required to show the matching history.
```
jpmedley, line 21:
```suggestion
`"declarativeNetRequestFeedback"` permission is required to show the matching history.
```

functional-samples/sample.dnr-rule-manager/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Running this extension
```
jpmedley, line 23:
```suggestion
## Run this extension
```

jpmedley: I can't remember how other samples phrase this. My suggestion is required by the corporate style guide. Ignore it if it's not consistent with other READMEs.

jpmedley (changes requested):

functional-samples/sample.dnr-rule-manager/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+## Overview
+
+This extension has two default rules:
+
+* Any main frame or XHR requests ending with `?nocookies=1` will have their cookies removed (URL Filter).
+
+* Any main frame or XHR requests matching `.*\.google\.com` will have their cookies removed (Regex Filter).
+
+For example, install this extension and try navigating to <https://github.com/GoogleChrome/chrome-extensions-samples?no-cookies=1> or <https://www.google.com> You should appear signed out. The number of requests modified by this extension will be displayed on the extension's badge.
```
oliverdunk, line 13:
```suggestion
For example, install this extension and try navigating to <https://github.com/GoogleChrome/chrome-extensions-samples?no-cookies=1> or <https://www.google.com> - you should appear signed out. The number of requests modified by this extension will be displayed on the extension's badge.
```

functional-samples/sample.dnr-rule-manager/manifest.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "no-cookies Rule Manager",
+  "version": "0.1",
+  "manifest_version": 3,
+  "description": "Uses the chrome.declarativeNetRequest API to remove the \"Cookie\" header from requests matching custom url filters.",
```
oliverdunk, line 5: Given this is used by the samples page, I think we should say more about how this differs from the other sample (e.g gives you a way to manage rules)?

functional-samples/sample.dnr-rule-manager/manager.js
oliverdunk: I think there's too much nesting/use of parameters as variables here which is confusing.

How about having:

```
function appendRuleItem(container, initialValues) {
```

Then, rather than updating `initialValues` when something changes, we can just use `getCurrentRuleValues` whenever we need the values and also update it to return the ID if needed. Does that make sense?

functional-samples/sample.dnr-rule-manager/manager.js
```diff
@@ -0,0 +1,199 @@
...
+    conditionValue: '',
+    caseSensitive: true
+  });
+});
+
+viewRuleButton.addEventListener('click', async () => {
+  const rules = await chrome.declarativeNetRequest.getDynamicRules();
+  const rulesString = JSON.stringify(rules, null, 2);
+  const newWindow = window.open();
+  newWindow.document.write(
```
oliverdunk, line 31: Can we find a way to do this by constructing an element vs. writing HTML (similar to wanting to avoid innerHTML)?

functional-samples/sample.dnr-rule-manager/manager.js
```diff
@@ -0,0 +1,199 @@
...
+    setSaveButtonDisabled();
+    // Update parameter
+    id = _id;
+    conditionType = _conditionType;
+    conditionValue = _conditionValue;
+    caseSensitive = _caseSensitive;
+  });
+
+  // Disable the save button
+  function setSaveButtonEnabled() {
```
oliverdunk, line 99: Can we have `setSaveButtonEnabled(enabled) {` to avoid two functions?

functional-samples/sample.dnr-rule-manager/manager.js
```diff
@@ -0,0 +1,199 @@
...
+    ruleItem.querySelector('.save-rule').disabled = false;
+  }
+  function setSaveButtonDisabled() {
+    ruleItem.querySelector('.save-rule').disabled = true;
+  }
+  setSaveButtonDisabled();
+
+  // Set input change handlers
+  async function verify() {
```
oliverdunk, line 108: Can we try to remove some nesting, e.g have verify take an element and live outside of `appendRuleItem`?

oliverdunk (changes requested):
Left some additional thoughts :)

jpmedley (approved):
LGTM for README only

oliverdunk (approved):
LGTM! Thanks for all of the work on this, really cool given it's entirely new 👏

PR #1023, opened 2023-10-27
Add API sample for User Scripts API

oliverdunk:
Adds a new API sample for the User Scripts API.

The feature is not currently launched, so there are a few caveats for testing (that will be resolved before merge):

- You need to test on **Chromium**, and use the following flag: `--enable-features="ApiUserScripts,RestrictDeveloperModeAPIs"`.
- The code option will register correctly but never be executed.

## Screenshots

|Developer mode disabled|Developer mode enabled|
|---|---|
|Screenshot 2023-10-27 at 14 43 46|Screenshot 2023-10-27 at 14 44 20|


api-samples/userScripts/options.js
```diff
@@ -0,0 +1,108 @@
...
+  const type = FORM.elements[TYPE_INPUT_NAME].value;
+  const script = FORM.elements[SCRIPT_TEXTAREA_NAME].value;
+
+  // Save to storage.
+  chrome.storage.local.set({
+    type,
+    script
+  });
+
+  // Unregister existing script.
```
sebastianbenz, line 70: unregister/register feels like such a common case it might make sense to make this more copy pasteable by making the function more generic, e.g.:

```
const userScript = {
  id: ...,
  matchPattern: ...,
  file: ...,
  script: ...,
}
registerUserScript(userScript)
```

Might be worth adding this method also to the docs cc @jpmedley

oliverdunk: Good callout. There is actually a `chrome.userScripts.update` which we can also look at including. It wasn't in the old copy of the docs I was looking at but looks like it's available now :)

api-samples/userScripts/README.md
```diff
@@ -0,0 +1,29 @@
+# chrome.userScripts API
+
+This sample demonstrates using the [`chrome.userScripts`](https://developer.chrome.com/docs/extensions/reference/scripting/) API to inject JavaScript into web pages.
+
+## Overview
+
+Once this extension is installed, clicking this extension's action icon will open an options page.
```
jpmedley, line 7:
```suggestion
Clicking this extension's action icon opens an options page.
```

jpmedley (changes requested):

api-samples/userScripts/options.js
jpmedley: https://source.chromium.org/chromium/chromium/src/+/main:extensions/common/api/user_scripts.idl;l=102

Already in the source.

api-samples/userScripts/README.md
```diff
@@ -0,0 +1,29 @@
...
+## Features
+
+This sample allows you to inject the following:
+
+- Files
+- Arbitrary code
+
+## Implementation Notes
+
+The User Scripts API requires developer mode. We check for this by attempting to access `chrome.userScripts`, which throws an error on property access if it is disabled.
```
AmySteam, line 27:
```suggestion
The User Scripts API requires users to have developer mode enabled. We check for this by attempting to access `chrome.userScripts`, which throws an error on property access if it is disabled.
```
This should be more explicit. Right now, I need clarification on whether this will be needed during production.

jpmedley: +1

api-samples/userScripts/options.js
oliverdunk: Switched to using this :)

api-samples/userScripts/manifest.json
```diff
@@ -0,0 +1,16 @@
+{
+  "name": "User Scripts API Demo",
```
AmySteam, line 2: @oliverdunk Sorry forgot to ask, can we add a minimum_chrome_version?

api-samples/userScripts/README.md
```diff
@@ -0,0 +1,29 @@
...
+## Features
+
+This sample allows you to inject the following:
+
+- Files
+- Arbitrary code
+
+## Implementation Notes
+
+The User Scripts API requires users to have developer mode enabled. We check for this by attempting to access `chrome.userScripts`, which throws an error on property access if it is disabled.
```
jpmedley, line 27:
```suggestion
The User Scripts API requires users to enabled developer mode. We check for this by attempting to access `chrome.userScripts`, which throws an error on property access if it is disabled.
```

jpmedley: Can you do feature detection as you would in a web page?

```
if (userScripts in chrome) { ... }
```

api-samples/userScripts/README.md
```diff
@@ -0,0 +1,29 @@
...
+
+When a change is made on the options page, we use the `chrome.userScripts` API to update the user script registration.
```
jpmedley, line 29:
```suggestion
When a change is made on the options page, use the `chrome.userScripts` API to update the user script registration.
```

jpmedley (changes requested):

api-samples/userScripts/manifest.json
oliverdunk: Done!

api-samples/userScripts/README.md
oliverdunk: Unfortunately not. The property always exists, even if developer mode is disabled.

jpmedley (approved):

PR #1030, opened 2023-11-15
Make the README and manifest consistent.

jpmedley:
No bug. I need to reference this in new content, so it needs to be consistent.


oliverdunk (approved):
LGTM, thanks for the fix.

PR #1039, opened 2023-12-01
Fix broken "All Frames" option in api-samples/scripting

zyv:
Apparently a copy & paste error.


JokerHarley89 (approved):

oliverdunk (approved):
Thanks for spotting this and opening a fix!

PR #1040, opened 2023-12-01
Create SECURITY.md

vovachello:


oliverdunk:
Closing this. If you'd like to add a new file, please open an issue and we can discuss what would be needed. I think for something like this we would want to agree on the desire and then make sure the template is actually filled out in the PR.

vovachello:
~~>

PR #1052, opened 2023-12-08
Update popup.js

hugolpz:
Just clean up.


sebastianbenz:
Thanks, but the formatting is correct (we use prettier). To check run `npm run lint`

PR #1056, opened 2023-12-20
Add offscreen user media sample

### LINKED ISSUE #821: MV3 user's mic ans cam permissions using iframe
I am building a chrome extension that records user's cam, mic, and screen.

For the screen, I think I took care of it by using desktopCapture API in background.js. 

The problem comes in for user's cam and mic.

I have an iframe injected into the shadow dom. From that iframe, I request for the permissions of the use of the user's cam and mic by using `"navigator.mediaDevices.getUserMedia({audio: true, video:true})"`. The reason for requesting mic and cam permissions inside the iframe is because I want to ask for the permissions on behalf of my extension app `(eg -> "[myExtension] wants to Use" instead of "www.google.com wants to use your camera and microphone")` and hold onto the permissions across all tabs, only asking once for the permissions.

I want to display user's cam and mic streams in content script, not inside the iframe. In order for me to stream the user's cam and mic in content script, somehow I need to find a way to get the streams of the cam and mic from the iframe and send it over to content script and display the streams. I believe this is where chrome.tabCapture API comes in. 

However, in manifest version 3, I cannot use chrome.tabCapture in backgrund.js (now also known as service workser). I think that's causing a major issue in this case. 

As I researched more into this specific case, I've found that there's a new feature called, offscreen API.

Can I use offscreen API to achieve what I'm trying to achieve?
- `requesting permission from the permission.html inside the iframe, which will be created using offscreen API`
- `use that permission in content script to use navigator.mediaDevices.getUserMedia() to get the streams of the user's cam and mic.`

Is this even possible in manifest version 3 currently? 

Here are the references I've gone over to solve this issue : 
https://bugs.chromium.org/p/chromium/issues/detail?id=1214847
https://bugs.chromium.org/p/chromium/issues/detail?id=1339382
https://stackoverflow.com/questions/74773408/chrome-tabcapture-unavailable-in-mv3s-new-offscreen-api
https://github.com/GoogleChrome/chrome-extensions-samples/issues/627
https://github.com/w3c/webextensions/issues/170
https://groups.google.com/a/chromium.org/g/chromium-extensions/c/ffI0iNd79oo/m/Dnoj6ZIoBQAJ?utm_medium=email&utm_source=footer

aakash232:
This recipe shows how to use User Media in an Extension Service Worker using the Offscreen document.

The extension aims to capture audio from the user in the context of extension (globally) via a media recorder in the offscreen document.

Aims to add a new sample fixing the issue : [MV3 user's mic and cam permissions using iframe](https://github.com/GoogleChrome/chrome-extensions-samples/issues/821) 
Fixes #821


aakash232:
@patrickkettner 
As discussed on the mail, raised the PR for the sample.
**Please review the same.**

aakash232:
Hi @oliverdunk 
Can you please take a look on this and guide me ahead?
Thanks

spikecodes (approved):

aakash232:
@oliverdunk 
As discussed on the mail, raised the PR for the sample.
Please review and guide me for the same.

functional-samples/cookbook.offscreen-user-media/manifest.json
```diff
@@ -0,0 +1,25 @@
...
+  "version": "1.0",
+  "manifest_version": 3,
+  "description": "Shows how to record audio in a chrome extension via offscreen document.",
+  "background": {
+    "service_worker": "background.js"
+  },
+  "content_scripts": [
+    {
+      "matches": ["<all_urls>"],
+      "js": ["contentScript.js"]
```
sorokinvj, line 12: you can do just this 
```suggestion
      "js": ["contentScript.js", "requestPermissions.js"]
```
and run it without all the fuss around messaging to offscreen and back

sorokinvj: however I still get permission in `prompt` state, so I am probably missing something from your code

aakash232: @sorokinvj 

#821  : The reason for requesting mic and cam permissions inside the iframe is because we want to ask for the permissions on behalf of the extension app (`eg -> "[myExtension] wants to Use" instead of "www.google.com wants to use your camera and microphone"`) and hold onto the permissions across all tabs, only asking once for the permissions.

---------------------------------

Your suggestion of injecting CS directly via manifest will not help in accomplishing above goal. But will definitely ease around the fuss if you don't need it in the context of your extension.


Attaching a snapshot for reference how your suggested code will ask for permissions.

image

Let me know if I am missing out something in your proposed flow.
Thanks.

patrickkettner:
hi @aakash232, I am really sorry this never got a proper review. your PR proved out the core idea here (grant mic to the extension origin from a page, record in the offscreen document) and #1717 builds on exactly that, in a smaller shape that avoids the all_urls content script. closing in favor of that one, and thank you for working out the hard part.

PR #1058, opened 2023-12-29
Fix 'invalid assignment to const'

mingpepe:


oliverdunk:
Thanks so much for the contribution! It looks like we actually merged [another fix](https://github.com/GoogleChrome/chrome-extensions-samples/pull/1067), so I don't think this is needed anymore. Sorry for that but thanks again, the work is still just as appreciated.

Issue #1054, opened 2023-12-19
code from function-sample-tutorial.quick-api-reference encounters authentication issue with API: Error Code 401


ahauser16:
I followed the Google Chrome tutorial located here: [developer.chrome.com/.../tutorial/service-worker-events](https://developer.chrome.com/docs/extensions/get-started/tutorial/service-worker-events).  I uploaded the official folder from Google Chrome's functional-samples folder and did not use my own code because I thought it was my mistake that caused the error.  I successfully uploaded the extension, typed "api" into the address bar, hit the TAB key and typed "runtime".  A new tab opened to the Runtime API documentation as expected, however, I did not see the "TIP" icon anywhere on the page.  I looked in the console and found two errors:

error one:
Uncaught (in promise) TypeError: Cannot read properties of null (reading 'append')
    at content.js:25:7

error two:
Failed to load resource: the server responded with a status of 404 ()
```
{
  "error": {
    "code": 401,
    "message": "API keys are not supported by this API. Expected OAuth2 access token or other authentication credentials that assert a principal. See https://cloud.google.com/docs/authentication",
    "status": "UNAUTHENTICATED",
    "details": [
      {
        "@type": "type.googleapis.com/google.rpc.ErrorInfo",
        "reason": "CREDENTIALS_MISSING",
        "domain": "googleapis.com",
        "metadata": {
          "method": "google.internal.developerprofiles.v1.profile.ProfileService.GetProfile",
          "service": "developerprofiles-pa.googleapis.com"
        }
      }
    ]
  }
}
```

**To Reproduce**
Steps to reproduce the behavior, or file the issue is found in:

1. Follow tutorial located here: 'https://developer.chrome.com/docs/extensions/get-started/tutorial/service-worker-events'
2. Load unpacked extension folder from Google Chrome's github located here: 'https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/functional-samples/tutorial.quick-api-reference'
3. Once extension is successfully uploaded type "api", hit "TAB" key as instructed by tutorial, enter "runtime" and hit "ENTER" key.
4. A new tab opens to the Runtime API documentation.
5. At this point in the tutorial (https://developer.chrome.com/docs/extensions/get-started/tutorial/service-worker-events#open-tip') you are instructed to "Open Tip of the Day" but it does not appear on the screen or in the DOM.
7. See error below listed in the ERRORS tab of the chrome://extensions page:
[...]

PR #1059, opened 2023-12-31
Bug #1054 resolved

itzmeimran:
A quick fix of issue


oliverdunk:
Thanks for taking a look at this. Have you checked to see if the tip appears anywhere now? I suspect with your change we might just have moved to a silent error unfortunately.

patrickkettner:
hi @itzmeimran, sorry for the wait. this has since been fixed on main, so closing. thanks for sending it in though!

PR #1060, opened 2024-01-01
add chrome.power/keep awake as an api sample

patrickkettner:
This is an [existing
extension](https://chromewebstore.google.com/detail/keep-awake/bijihlabcfdnabacffofojgmehjdielb)
that was previously hosted on chromium.org. It has ~400k users, but had
been orphaned over the years.  I took it over to migrate to Manifest v3.
Given we are hosting the code for the Font Settings extension, and we
don't have a chrome.power sample, I thought it made sense to host the
code here.


api-samples/power/README.md
```diff
@@ -0,0 +1,13 @@
+# chrome.power
+
+This extension demonstrates the `chrome.power` API by allowing users to override their system's power management features.
+
+## Overview
+
+The extension creates add a popup, that cycles different states when clicked. It will go though a mode that prevents the display from dimming or going to sleep, a mode that keeps the system awake but allows the screen to dim/go to sleep, and whatever the system's default is.
```
oliverdunk, line 7:
```suggestion
The extension adds a popup that cycles different states when clicked. It will go though a mode that prevents the display from dimming or going to sleep, a mode that keeps the system awake but allows the screen to dim/go to sleep, and a mode that uses the system's default.
```

api-samples/power/README.md
```diff
@@ -0,0 +1,13 @@
...
+
+## Running this extension
+
+1. Clone this repository.
+2. Load this directory in Chrome as an [unpacked extension](https://developer.chrome.com/docs/extensions/mv3/getstarted/development-basics/#load-unpacked).
+3. Pin the extension to the taskbar and click the action button.
```
oliverdunk, line 13:
```suggestion
3. Pin the extension and click the action button.
```

api-samples/power/manifest.json
```diff
@@ -0,0 +1,26 @@
+{
+  "manifest_version": 3,
+
+  "name": "__MSG_extensionName__",
```
oliverdunk, line 4: Unfortunately the sample list generator doesn't have support for this, so I think we need to either add that or directly include the name and description here.

oliverdunk (changes requested):
Left a few comments, let me know when you've tweaked and I can test and approve :)

api-samples/power/manifest.json
patrickkettner: Added support for it in cdde058f49028f4d9fcb7c0ab62256deb230bd7d

oliverdunk: Thanks! I'll leave a couple of comments but overall looks good :)

.repo/sample-list-generator/src/utils/manifest.ts
```diff
@@ -1,11 +1,32 @@
...
+import { ManifestData, LocaleData } from '../types';
+const localeRegex = /__MSG_([^_]*)__/
 
 export const getManifest = async (
   manifestPath: string
 ): Promise<ManifestData> => {
   const manifest = await fs.readFile(manifestPath, 'utf8');
   const parsedManifest = JSON.parse(manifest);
 
+  if (manifest.includes('__MSG_')) {
```
oliverdunk, line 12: It makes me really uncomfortable searching through an unparsed file like this 😅  - would you be open to removing the if here, and just move loading of the messages.json file to when we find a value that starts with MSG? It does mean we might load the file multiple times unless you want to do some more tweaking but I don't think that's the end of the world.

oliverdunk (changes requested):
Left one more comment, otherwise I tested this and it looks good so I'm happy.

.repo/sample-list-generator/src/utils/manifest.ts
patrickkettner: abstracted in bd6bf32b0b4b1b101899937aae785d654efcf36b.

For the sake of localization on dcc,  we could potentially localize the metadata for the samples extension and update this to take a langauge flag as an argument to load that language, rather than the currently hardcoded english.

PR #1067, opened 2024-01-09
Fix: use let to prevent reassignment error for sandbox result

stephenkao:
In the sandbox/sandbox example, a variable `result` is created as a const, but it's attempted to be reassigned later in the code.  This changeset just changes the `result` variable to be a reassignable variable instead.

Screenshot 2024-01-09 at 18 32 15


sebastianbenz (approved):
Thanks for fixing!

PR #1075, opened 2024-01-27
Add Dependabot

RyanLua:
Adds Dependabot to update GitHub Actions and NPM.


oliverdunk (approved):
Thanks for opening this! We wouldn't usually accept PRs changing repository configuration without discussion, since usually these are somewhat subjective. This seems like a reasonable change though so going to merge.

PR #1080, opened 2024-02-08
Fix direction of the keys.

jpmedley:
This sample cycles between tabs using key combinations that include arrow keys. This had a bug wherein the right arrow circled left and the left arrow circled right. This PR fixes that.


oliverdunk (approved):
Oh, nice spot! Thanks for this.

PR #1083, opened 2024-02-15
update DNR samples to work in new DCC

### LINKED ISSUE #1082: Manifest V3 declarativeNetRequest redirect sample is completely broken on latest chrome 121
⚠️ If you have general Chrome Extensions questions, consider posting to the [Chromium Extensions Group](https://groups.google.com/a/chromium.org/forum/#!forum/chromium-extensions) or [Stack Overflow](https://stackoverflow.com/questions/tagged/google-chrome-extension).

**Describe the bug**
The declarativeNetRequest redirect does not do anything, sevice worker doesn't react and the browser behaves as if there's no extension at all, not redirecting anywhere.

**To Reproduce**
Steps to reproduce the behavior, or file the issue is found in:
1. Get code from https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/api-samples/declarativeNetRequest/url-redirect
2. Load as unpacked extension in Chrome 121 retail
3. Go to one of the example websites listed such as: https://developer.chrome.com/docs/extensions/mv2/
4. The website loads and nothing happens.

**Expected behavior**
Should redirect to:
https://developer.chrome.com/docs/extensions/mv3/intro/

**Notes**
I believe that Manifest V3 declarativeNetRequest API is currently completely broken and inactive.
Maybe it's a bug with Chrome itself?
Tested on chrome 121.0.6167.185

patrickkettner:
fixes #1082

### DIFF
--- a/api-samples/declarativeNetRequest/url-redirect/README.md
+++ b/api-samples/declarativeNetRequest/url-redirect/README.md
@@ -15,3 +15,5 @@ Once this extension is installed, any requests made in the main frame to the fol
 ## Implementation Notes
 
 This sample uses the `chrome.declarativeNetRequest.onRuleMatchedDebug` event which is only available in unpacked extensions.
+
+The rules match the `xmlhttprequest` resource type in addition to `main_frame`. For returning visitors, developer.chrome.com serves navigations through its service worker, so no `main_frame` network request is made; the request issued by the site's service worker is matched instead.
--- a/api-samples/declarativeNetRequest/url-redirect/rules_1.json
+++ b/api-samples/declarativeNetRequest/url-redirect/rules_1.json
@@ -9,7 +9,7 @@
       }
     },
     "condition": {
-      "urlFilter": "https://developer.chrome.com/docs/extensions/mv2/",
+      "urlFilter": "https://developer.chrome.com/docs/extensions/mv2",
       "resourceTypes": ["main_frame"]
     }
   },
@@ -23,7 +23,7 @@
       }
     },
     "condition": {
-      "urlFilter": "https://developer.chrome.com/docs/extensions/reference/browserAction/",
+      "urlFilter": "https://developer.chrome.com/docs/extensions/reference/browserAction",
       "resourceTypes": ["main_frame"]
     }
   },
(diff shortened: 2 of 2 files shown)


oliverdunk (approved):
LGTM, thanks for the fix!

hrmtsh2:
Hi, I came here from the discussion on #1082 as I was having the same issue. However this fix also doesn't seem to change things, and somewhere down the line the changes were reverted?
I was trying to run this sample (in relation to my potential work on building an interactive tool to view changes in rules for DNR-bassed extensions for the GSoC program), but it doesn't seem to work.
As mentioned in #1082 by @oliverdunk, I tried to keep both `"main_frame"`, `"xmlhttprequest"` in the `resourceTypes` key to accommodate both sites with service workers and with main_frame, but to little avail. The only change I see is that the console message in `service_worker.js` now shows the message confirming the redirection (I assume the tabId of -1 indicates that it is not tab-specific?).
However, the redirection still doesn't occur. Kindly tell me where I am going wrong.

patrickkettner:
@oliverdunk finally got back to this. I dug into why the reports here were inconsistent: developer.chrome.com serves navigations through its service worker for returning visitors, so there is no `main_frame` network request for DNR to match. Fresh profiles redirect fine, returning visitors get nothing, which lines up with what @hrmtsh2 saw (rule match logged with tabId -1, no actual redirect).

Matching `xmlhttprequest` as well (your suggestion from #1082) fixes it on current Chrome. I tested this branch on 149 with a fresh profile and again with the site's service worker registered, and both redirect now. Also updated the redirect destinations since the old ones have themselves become 301s, and added a README note about the service worker behavior.

Given the sample now needs that caveat explained to make sense, does it still seem worth merging, or would you rather we abandon it?

oliverdunk:
> Given the sample now needs that caveat explained to make sense, does it still seem worth merging, or would you rather we abandon it?

I'm still happy to land this given it is a simple change and gets the sample working more reliably. However, I think this is a confusing sample. Longer term my preference would be to just redirect some requests to example.com or similar. What do you think?

patrickkettner:
fine with me. ill assign a issue for me to redo the same soon.

Issue #1020, opened 2023-10-21
Dictionary side panel example doesn't work with Uncaught (in promise) Error: Could not establish connection. Receiving end does not exist. 


AlexMikhalev:
**Describe the bug**
Dictionary side panel example doesn't work with the error
```
Uncaught (in promise) Error: Could not establish connection. Receiving end does not exist. 
``` 
in the service worker.
**To Reproduce**

1. Clone chrome-extension-samples, commit b25a56bf8ea0bf339c0f0b5e882d5695e32726f7 
2. cd functional-samples/sample.sidepanel-dictionary 
3. Follow the steps in Readme to install the extension
4. Define item appear in context menu
5. Clicking on define sidepanel will not appear 
6. Service worker will have error:
"Uncaught (in promise) Error: Could not establish connection. Receiving end does not exist."

Chrome:
Version 118.0.5993.70 (Official Build) (64-bit)
No other extensions or blockers are installed or active.

**Expected behavior**
Sidepanel appear

AmySteam:
Hey @AlexMikhalev, thanks for flagging this. 😃

Out of curiosity, did you select the word "extensions", then right-click and choose the context menu Define? 

This sample has a long list of steps, it can be easily missed 😟

AlexMikhalev:
Yes. The point of a bug report is that we need working examples of how to pass data - even the string into the side panel with any reliability that our users will have it working in their environment. The case of nothing selected shall be handled too.

ChuckStrusz:
> The fix is to inject client-side script with chrome.runtime.sendMessage

As someone who is floundering with the absolute paucity of good (ie, functioning) examples of V3 extensions, I'm hoping you can offer a bit more detail on this fix for this broken sample (I have the same issues) - is there any chance you could post the full updated source for the file in question, or indicate which file you mean by "client-side script" since I think there are two - both service-worker.js and sidepanel.js are "client-side", right?

ChuckStrusz:
I am having this same issue, but I'll note that the instructions in the README, as written, _do not work_.  To wit, Step 3 is "Open the side panel UI" - but you **can't** open the side panel UI, because the extension button is disabled (greyed out).  I have tried this in a normal and an incognito window with the same results.
![extension](https://github.com/GoogleChrome/chrome-extensions-samples/assets/22245534/b9cbced5-f2a2-44c5-8e52-b42c868b9ea6)

patrickkettner:
Hi chuck
[...]

PR #1087, opened 2024-02-19
Update dictionary sample to use sidePanel.open()

oliverdunk:
Updates the dictionary side panel example to use the sidePanel.open() API. This means we are also able to demonstrate a way of safely passing data to the side panel without hitting any race conditions.

Fixes https://github.com/GoogleChrome/chrome-extensions-samples/issues/1020

### DIFF
--- a/functional-samples/sample.sidepanel-dictionary/sidepanel.js
+++ b/functional-samples/sample.sidepanel-dictionary/sidepanel.js
@@ -24,9 +24,13 @@ chrome.storage.session.get('lastWord', ({ lastWord }) => {
 });
 
 chrome.storage.session.onChanged.addListener((changes) => {
-  if ('lastWord' in changes) {
-    updateDefinition(changes['lastWord'].newValue);
+  const lastWordChange = changes['lastWord'];
+
+  if (!lastWordChange) {
+    return;
   }
+
+  updateDefinition(lastWordChange.newValue);
 });
 
 function updateDefinition(word) {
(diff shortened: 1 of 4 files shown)


functional-samples/sample.sidepanel-dictionary/sidepanel.js
```diff
@@ -19,14 +19,27 @@ const words = {
...
+  updateDefinition(lastWord);
+});
 
-    // Show word and definition.
-    document.body.querySelector('#definition-word').innerText = data.value;
-    document.body.querySelector('#definition-text').innerText =
-      words[data.value.toLowerCase()];
+chrome.storage.session.onChanged.addListener((changes) => {
+  if ('lastWord' in changes) {
+    updateDefinition(changes['lastWord'].newValue);
```
sebastianbenz, line 28: optional: I prefer avoiding using the string literal twice

```
const lastWord = changes['lastWord'];
if (!lastWord) {
  return
}
updateDefinition(lastWord.newValue);
```

sebastianbenz (approved):
Nice update! That's a good pattern to demo.

functional-samples/sample.sidepanel-dictionary/sidepanel.js
oliverdunk: Gave a change a try, let me know if that looks better.

sebastianbenz: I think it looks better :-) - thanks!

PR #1117, opened 2024-03-02
add devDep @types/chrome

SherlockHomer:
Hi there, let's use devDep `types/chrome` to code easily.
image

> Tip: vscode plug-in Chrome Extension API is not updated in time

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,6 +9,7 @@
       "version": "1.0.0",
       "license": "Apache 2.0",
       "devDependencies": {
+        "@types/chrome": "^0.0.262",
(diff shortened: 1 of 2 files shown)


oliverdunk (changes requested):
Thanks, this is a great idea! Would you be open to using [`chrome-types`](https://www.npmjs.com/package/chrome-types), which we publish from the Chromium source and has all APIs? We don't have the perfect story for when to pick which yet but I think that makes the most sense for samples where we want to be writing things using new features.

SherlockHomer:
Why not? To be writing things using new features makes sense. 👍

PR #1118, opened 2024-03-02
Delete _archive/mv2/api/i18n directory

masalati1122:


patrickkettner:
Thanks for the PR @masalati1122 - what is the reasoning behind it, though? Why are you removing it?

PR #1134, opened 2024-03-27
native-messaging example support non-ascii language

iamqiz:
Update native-messaging-example-host to support non-ascii language


upperctrl (approved):

dkocich (commented):
The fix is correct — Python 3's `sys.stdout.write()` can't write bytes, and the native messaging protocol requires byte-length in the struct pack. Encoding to UTF-8 first and writing to `sys.stdout.buffer` is the right approach.

Minor style nit: PEP 8 recommends spaces around `=` in assignments (`message = message.encode('utf-8')`), but functionally this is fine.

PR #1141, opened 2024-04-06
Fix Time Unit Error in History Samples

### LINKED ISSUE #1140: Incorrect Time Unit in History API Sample Code
**Describe the bug**
In the [showHistory](https://github.com/GoogleChrome/chrome-extensions-samples/blob/ab934c0e6c6d0c5e8bc08a87f34f833df12e05ec/api-samples/history/showHistory/popup.js#L49C1-L52C63) sample, the time unit used for fetching the history is incorrect. 
``` js
// To look for history items visited in the last week,
// subtract a week of microseconds from the current time.
let microsecondsPerWeek = 1000 * 60 * 60 * 24 * 7;
let oneWeekAgo = new Date().getTime() - microsecondsPerWeek;
```

**Expected behavior**
The time unit should be in millisecond unit and not microsecond.

**Notes**
I would like to fix this error.

atick-faisal:
Fixes #1140

### DIFF
--- a/api-samples/history/showHistory/popup.js
+++ b/api-samples/history/showHistory/popup.js
@@ -47,9 +47,9 @@ function buildPopupDom(divName, data) {
 // and show those links in a popup.
 function buildTypedUrlList(divName) {
   // To look for history items visited in the last week,
-  // subtract a week of microseconds from the current time.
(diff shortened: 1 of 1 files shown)


oliverdunk (approved):
Good spot! Thanks for the fix.

PR #1149, opened 2024-04-26
Upgrade to ESLint 9

oliverdunk:
ESLint has undergone a large migration to a new "flat" config structure.

I followed the migration guide here to migrate: https://eslint.org/docs/latest/use/configure/migration-guide

A couple of notes:

- I removed `eslint-plugin-import`, which isn't currently compatible with ESLint 9 and doesn't expect to have support soon. We don't use imports often so I think this is the best call, but open to thoughts.
- I've ignored the .repo folder for now which had some errors - we should fix this in a follow-up PR.


package.json
```diff
@@ -21,12 +21,13 @@
   },
   "homepage": "https://github.com/GoogleChrome/chrome-extensions-samples#readme",
   "devDependencies": {
+    "@eslint/js": "^9.1.1",
```
sebastianbenz, line 24: We should pin versions consitently

sebastianbenz (approved):

package.json
oliverdunk: Good idea - fixed!

PR #1264, opened 2024-08-26
Migrate to new prompt API

sebastianbenz:
...also add a check whether the browser supports the prompt API.


oliverdunk (approved):
Even simpler :)

PR #1347, opened 2024-11-12
Update Geminin nano samples

sebastianbenz:
* Link to latest articles
* Add systemprompt to prompt sample to provide better guidance on simple prompts such as 'hello'.
* Limit default topK to 3 (the default is currently set too high, this will change).
* Fix setup instructions in the readmes.


andreban (approved):
LGTM

PR #1350, opened 2024-11-16
Rename `privacy,txt` to `privacy.txt`

MahadMuhammad:
A slight fix on using `,` instead of `.`


oliverdunk (approved):
Thanks for spotting this and opening a PR!

PR #1372, opened 2024-12-13
Add environment variable for configuring storage bucket

oliverdunk:
Adds the STORAGE_BUCKET environment variable which is used for determining which bucket the prepare-chrome-types script loads type data from.

### DIFF
--- a/.repo/sample-list-generator/src/prepare-chrome-types.ts
+++ b/.repo/sample-list-generator/src/prepare-chrome-types.ts
@@ -4,12 +4,19 @@ import fs from 'fs/promises';
 import { ExtensionApiMap } from './types';
 import { ReflectionKind } from 'typedoc';
 
+// Bucket used to store processed types data
+const STORAGE_BUCKET = process.env.STORAGE_BUCKET;
+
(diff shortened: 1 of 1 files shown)


.repo/sample-list-generator/src/prepare-chrome-types.ts
```diff
@@ -4,12 +4,19 @@ import fs from 'fs/promises';
 import { ExtensionApiMap } from './types';
 import { ReflectionKind } from 'typedoc';
 
+// Bucket used to store processed types data
+const STORAGE_BUCKET = process.env.STORAGE_BUCKET;
```
sebastianbenz, line 8: Should we mention this in the readme?

sebastianbenz (approved):

xximwon (approved):

.repo/sample-list-generator/src/prepare-chrome-types.ts
oliverdunk: Yeah. Once we figure out how we're going to handle this going forward I'll update it.

PR #1378, opened 2024-12-23
fix: the extension page will refresh after clicking the button. We need to add preventDefault to prevent form default behavior.

minggeorgelei:


oliverdunk (approved):
Good catch! Thanks for the fix :)

Issue #657, opened 2021-12-24
is there any methods to control the border radius of popup.html? 


laulliam:
is there any methods to control the border radius of popup.html?

compiled-krishan:
This is not possible in the current manifest versions. May be we can see updates on this in future from google.

hunterphillips:
Everything I've seen says its not possible, but it looks like the Google Keep extension is pulling it off (can see in the 3rd screenshot)

https://chrome.google.com/webstore/detail/google-keep-chrome-extens/lpcaedmchfhocbbapmcbpinfpgnhiddi?hl=en

patrickkettner:
Checking out [ the code](https://robwu.nl/crxviewer/?crx=https://chrome.google.com/webstore/detail/lpcaedmchfhocbbapmcbpinfpgnhiddi), you will see it is not using a popup. They are injecting an iframe into the page, and styling that.

Vedbhanushali:
@patrickkettner it is confusing I checked the code it seems like compiled code, can you share an example of achieving this or source code reference.

patrickkettner:
An example of what? They aren’t doing what the OP thought they were doing.

Vedbhanushali:
@patrickkettner  you mean that using content.js they are injecting iframe in current tab instead of using popup.html

NoamRivlin:
have mv3 made this possible or is it still not possible to control?

oliverdunk:
Hi @NoamRivlin, there aren't any changes here I'm afraid.

Visalan-H:
sad

iamOmarFaruk:
It's possible to ignore popup.html and inject a layout through background.js. I just👨🏻‍💻 did it.

`chrome.action.onClicked.addListener((tab) => { // Get the current tab ID
  const tabId = tab.id;
  
  // Inject the content script into the current tab
  chrome.scripting.executeScript({
    target: { tabId: tabId },
    files: ['content.js']
  });
  
  console.log(`Content script injected into tab ${tabId}`);
}); `

and write the popup layout and inject in the current page, its basically a injected dom element 😁 

<img width="563" alt="Image" src="https://github.com/user-attachments/assets/e63898fc-b026-42b4-a5ce-cd7daef8c500" />

Obsessive-Curiosity:
thanks!

PR #1421, opened 2025-03-04
Add rounded iframe sample

ttw225:
This PR adds a new sample `sample.rounded-iframe`, which demonstrates how to use an iframe with rounded corners in a Chrome Extension.

> The inspiration for this template stems from discussions across various platforms, including [Stack Overflow](https://stackoverflow.com/questions/27899635/how-to-make-border-radius-in-popup-chrome-extension) and the [Chrome Extensions Samples repository](https://github.com/GoogleChrome/chrome-extensions-samples/issues/657), where developers have explored methods to implement rounded corners in extensions. Notably, the Google Keep Chrome Extension has been recognized for its effective use of rounded corners. Inspired by its advanced techniques, this template simplifies and adapts those methods to assist developers in creating more visually appealing extension interfaces.


dkocich (commented):
Nice concept. A few issues to fix before merging:

**1. Incomplete URL guard** — \ misses other restricted contexts. Use:
\
**2. Global namespace pollution** — storing handlers on \ / \ leaks into the page's global scope. Use an \ instead:
\
**3. Missing newlines at end of files** — \, \, \, \ all lack a trailing newline (will fail Prettier lint).

**4. Minor** — \ missing \ is fine because \ covers it, but worth noting in the README for clarity.

ttw225:
@dkocich  Thanks a lot for the detailed review and suggestions!

I’ve addressed all three points you mentioned.

Please let me know if there’s anything else I should adjust.

PR #1434, opened 2025-03-20
Update api-samples/bookmarks

ljjlee:
The extension currently assumes that a bookmark folder with ID "1" exists, and that children can be created below it. This is not guaranteed to be true.

Update the chrome.bookmarks.create() call to instead leave parentId unspecified, which will select the "Other bookmarks" folder.


ljjlee:
@oliverdunk would you be able to take a look at this PR too? Hopefully self-explanatory, but let me know if not.

oliverdunk (approved):
This makes sense! Thanks for the simple and well scoped PR :)

PR #1435, opened 2025-03-20
Disambiguate account/local folders sample bookmarks extension

ljjlee:
This shows how to use the recently-added `folderType` and `syncing` bookmark node properties to distinguish top-level folders that may existing in both syncing and non-syncing storages.


ljjlee:
@oliverdunk would you have time to review this change to the functional bookmarks sample?

oliverdunk (approved):
Thanks for opening this!

PR #1454, opened 2025-04-07
Add getProfileUserInfo example to sample identity extension

ljjlee:
This adds an example for the [getProfileUserInfo()](https://developer.chrome.com/docs/extensions/reference/api/identity#method-getProfileUserInfo) method, printing the primary account's email address.


oliverdunk (approved):
Thanks as always!

PR #1456, opened 2025-04-11
Replaced clearbit API (deprecated) (#1453)

ZaifSenpai:
Replaced the usage of Clearbit’s Free Logo API (logo.clearbit.com), which is Being Discontinued (#1453)
with Google's favicon snatcher

### DIFF
--- a/_archive/mv2/extensions/optional_permissions/logic.js
+++ b/_archive/mv2/extensions/optional_permissions/logic.js
@@ -28,7 +28,7 @@ function createTop(){chrome.topSites.get(function(topSites) {
     let hostname = (new URL(site.url)).hostname;
     let image = document.createElement('img');
     image.title = site.title;
-    image.src = 'https://logo.clearbit.com/' + hostname;
+    image.src = 'https://www.google.com/s2/favicons?sz=128&domain_url=' + hostname;
     url.appendChild(image);
     div.appendChild(url);
     div.appendChild(tooltip);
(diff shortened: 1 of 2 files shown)


robin450916 (approved):

oliverdunk (changes requested):
Thanks for the PR! Could you take a look at using the built-in favicon support in extensions?

MV2: `chrome://favicon/https://example.com/`
MV3: https://developer.chrome.com/docs/extensions/how-to/ui/favicons

ZaifSenpai:
@oliverdunk Thank you for your response.
What we need here is high resolution site icon to show. The solutions you mentioned returns low resolution images.

• Using `chrome://favicon/` (MV2):

image


• Using [favicon](https://developer.chrome.com/docs/extensions/how-to/ui/favicons) (MV3):

image


• Using Google's favicon snatcher:

image

oliverdunk:
> What we need here is high resolution site icon to show.

Could we tweak the sample to use smaller icons? Since the focus of the code is demonstrating optional permissions, I don't think we need high resolution images and I'd prefer not to depend on another API which as far as I can tell is not publicly documented and might also go away in the future.

ZaifSenpai:
> > What we need here is high resolution site icon to show.
> 
> Could we tweak the sample to use smaller icons? Since the focus of the code is demonstrating optional permissions, I don't think we need high resolution images and I'd prefer not to depend on another API which as far as I can tell is not publicly documented and might also go away in the future.

sure. on it.

ZaifSenpai:
New design:

image

oliverdunk (approved):
Looks great! Thanks for your work on this and your patience waiting for a review.

PR #1496, opened 2025-06-19
Handle SPA navigations in reading time example

oliverdunk:
In the move to new infrastructure for https://developer.chrome.com/, the site became an SPA which can soft navigate outside of browser navigations. Therefore, we need to update the sample to watch for changes to content after the content script is injected.


sebastianbenz (approved):
Nice! Thanks for fixing this!

PR #1503, opened 2025-07-06
Update popup.js

ShreyanshJain105:
refactor: modernize alt text translator with class-based architecture

Restructure codebase from procedural to object-oriented approach with improved  error handling, comprehensive documentation, and better separation of concerns.


oliverdunk:
Assigning to @sebastianbenz who wrote this sample. We might not want to make this change as we try to keep our code as simple as possible, but he can make the final decision. Either way, thank you for the contribution!

oliverdunk:
Please also sign the CLA when you can which we will need before being able to merge.

sebastianbenz:
Thanks for the contribution. Oliver is right, the implementation was intentional to keep things simple. In the future I recommend creating an issue before creating a PR. Thanks again!

PR #1514, opened 2025-07-27
Use getContexts to manage offscreen document lifecycle (recommended from docs)

madaster97:
In the [documentation](https://developer.chrome.com/docs/extensions/reference/api/offscreen#maintain_the_lifecycle_of_an_offscreen_document) for maintaining an offscreen document, the recommendation is to do the heavy lifting in the parameters that can be passed to `getContexts`:
```js
const existingContexts = await chrome.runtime.getContexts({
  contextTypes: ['OFFSCREEN_DOCUMENT'],
  documentUrls: [offscreenUrl]
});
```

But in the three functional examples showing offscreen documents, we don't follow that recommendation:
```js
const matchedClients = await clients.matchAll();
for (const client of matchedClients) {
  if (client.url.endsWith(OFFSCREEN_DOCUMENT_PATH)) {
    return true;
  }
}
```

In this PR, I've made the above edit to the `offscreen-dom` cookbook, along with 1 other edit: I've removed the *check for document* that we do before trying to close the document. Instead, I've just put in a warning if our attempt to close the document fails:
```js
chrome.offscreen.closeDocument().catch(error => {
  console.warn(
    'Received an error while attempting to close an offscreen document, after receiving a message assumed to come from an offscreen document: %o', error)
});
```

Does this look OK? Should we add it to the other 2 offscreen examples? (clipboard-write and geolocation)

### DIFF
--- a/functional-samples/cookbook.offscreen-dom/background.js
+++ b/functional-samples/cookbook.offscreen-dom/background.js
@@ -21,17 +21,38 @@ chrome.action.onClicked.addListener(async () => {
   );
 });
 
-async function sendMessageToOffscreenDocument(type, data) {
-  // Create an offscreen document if one doesn't exist yet
-  if (!(await hasDocument())) {
-    await chrome.offscreen.createDocument({
+let creating; // A global promise to avoid concurrency issues
+async function setupOffscreenDocument(path) {
+  // Check all windows controlled by the service worker to see if one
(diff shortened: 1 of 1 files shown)


oliverdunk:
Closing this for now. If you'd like to push some changes, feel free to let us know and we can reopen.

PR #1515, opened 2025-07-28
[Bug] Fix syntax error

vinlim:
```
try {
    if (data.type != 'audio-scribe' || !data || !isValidUrl(data.objectUrl)) {
      return;
    }
    // Check if it's an audio file
    const audio = await fetch(data.objectUrl);
    content = await audio.blob();
    if (!content.type || !content.type.startsWidth('audio/')) {
      return;
    }
  } catch (e) {
    console.log(e);
  }
```

Fix `startsWidth` with `startsWith`

PR #1516, opened 2025-07-28
Fix prompt format for image description in prompt array

vinlim:
Fix the prompt format from
```
const prompt = [
    `Please provide a functional, objective description of the provided image in no more than around 30 words so that someone who could not see it would be able to imagine it. If possible, follow an “object-action-context” framework. The object is the main focus. The action describes what’s happening, usually what the object is doing. The context describes the surrounding environment. If there is text found in the image, do your best to transcribe the important bits, even if it extends the word count beyond 30 words. It should not contain quotation marks, as those tend to cause issues when rendered on the web. If there is no text found in the image, then there is no need to mention it. You should not begin the description with any variation of “The image”.`,
    { type: 'image', content: imageBitmap }
  ];
```
to updated requirement
```
const prompt = [{
    role: 'user',
    content: [
      {
        type: 'text',
        value: `Please provide a functional, objective description of the provided image in no more than around 30 words so that someone who could not see it would be able to imagine it. If possible, follow an “object-action-context” framework. The object is the main focus. The action describes what’s happening, usually what the object is doing. The context describes the surrounding environment. If there is text found in the image, do your best to transcribe the important bits, even if it extends the word count beyond 30 words. It should not contain quotation marks, as those tend to cause issues when rendered on the web. If there is no text found in the image, then there is no need to mention it. You should not begin the description with any variation of “The image”.`
      },
      { type: 'image', value: imageBitmap }
    ]
  }];
  return await session.prompt(prompt);
```

### DIFF
--- a/functional-samples/ai.gemini-on-device-alt-texter/background.js
+++ b/functional-samples/ai.gemini-on-device-alt-texter/background.js
@@ -19,16 +19,18 @@ async function generateAltText(imgSrc) {
   const imageBitmap = await createImageBitmap(blob);
 
   // Run the prompt
-  const prompt = [{
-    role: 'user',
-    content: [
-      {
-        type: 'text',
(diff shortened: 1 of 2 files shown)


sebastianbenz:
Thanks for the fix! Can you please sign the CLA?

vinlim:
No problem. Fixed the Author for CLA purpose. Thanks.

vinlim:
The CI/lint failure is resulted by https://github.com/GoogleChrome/chrome-extensions-samples/pull/1515. It can be fixed by merging https://github.com/GoogleChrome/chrome-extensions-samples/pull/1518 first.

sebastianbenz (approved):

sebastianbenz:
Thanks a lot!

PR #1536, opened 2025-08-20
Fix observer to target 'devsite-article' instead of 'devsite-content'

abdessamadbettal:


sebastianbenz:
Did you test this change? What is the problem that this addresses?

sebastianbenz:
Closing as the sample currently works as expected.

PR #1597, opened 2025-12-19
Add Prompt API trial tokens for multimodal extension samples

michaelwasserman:
@sebastianbenz, please take a look, thanks!
AFAICt, these aren't published on the CWS, so we don't need to update anything else?


oliverdunk (approved):
Thanks Mike! Looks great :)

PR #1610, opened 2026-01-08
security: restrict content script matches to specific domain in favicon sample

RinZ27:
The `sample.favicon-cs` currently uses `<all_urls>` for its content script matches.

Injecting scripts into every page visited by the user is an over-privileged practice that increases the extension's attack surface unnecessarily. Restricting matches to a specific domain (e.g., `https://www.google.com/*`) demonstrates the Principle of Least Privilege while still effectively showing how the Favicon API works in content scripts.

### DIFF
--- a/functional-samples/sample.favicon-cs/README.md
+++ b/functional-samples/sample.favicon-cs/README.md
@@ -1,6 +1,6 @@
 ## Fetching a favicon in a content script
 
-This example fetches the favicon from www.google.com and inserts it at the top left of every page.
+This example fetches the favicon from www.google.com and inserts it at the top left of Google search pages.
 
 Note: This extension does not work on `chrome://extensions`.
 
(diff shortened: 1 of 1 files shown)


RinZ27:
Updated the README to match the new host permissions. It now correctly points to Google pages instead of claiming to work everywhere.

romscipharm-blip:
Hi @RinZ27,

Thanks for your work on restricting content script matches to specific domains. From a user's perspective, it can still be challenging to quickly understand what a browser extension does and what data it accesses.

Just wanted to share this user perspective. Thanks!

RinZ27:
@romscipharm-blip Appreciate the feedback! I totally agree—technical security fixes are only one side of the coin. Real transparency comes from making those permissions easy for anyone to understand. Hopefully, these small refinements in the samples help set a better standard for that.

PR #1685, opened 2026-05-24
Fix side panel sample tab path setup

### LINKED ISSUE #1477: cookbook.sidepanel-open sample does not behave correctly
https://github.com/GoogleChrome/chrome-extensions-samples/blob/main/functional-samples/cookbook.sidepanel-open/service-worker.js doesn't really work as the author intended. 

The sample tries to create a button that will be appended to a page, the button will send a message, and a message listener will open the sidePanel to a secondary path instead of the primary path for the panel.

The problem is that the code does this:

```javascript
      await chrome.sidePanel.open({ tabId: sender.tab.id });
      await chrome.sidePanel.setOptions({
        tabId: sender.tab.id,
        path: 'sidepanel-tab.html',
        enabled: true
      });
```

Since open is done first the panel is opened to the primary path and then the options get set. If you press the button a second time the options that were previously set (including the sidepanel-tab.html path) will actually show.

What you really need to be able to do is to call setOptions followed by open but you can't. Trying to do that will cause the user gesture context to be lost or invalidated by the time the open call happens. This results in an  `Uncaught (in promise) Error: sidePanel.open() may only be called in response to a user gesture.`

I don't think the example works how the author intended which is to show a button added to a web page that results in the sidePanel being opened to a different path. The only way to make it work is to hit the button twice. The first time will show the panel and set the options and the second time will re-open the panel to the options set on the previous button press.

The example is misleading and should be changed.

KirtiRamchandani:
## Summary
- Configure the tab-specific side panel path before the extension page button can open the panel.
- Have the content script request tab-specific setup before enabling its injected button.
- Keep the click-triggered path focused on `chrome.sidePanel.open()` so the first click opens `sidepanel-tab.html`.

Fixes #1477.

## Testing
- `npx eslint functional-samples/cookbook.sidepanel-open/content-script.js functional-samples/cookbook.sidepanel-open/script.js functional-samples/cookbook.sidepanel-open/service-worker.js`

### DIFF
--- a/functional-samples/cookbook.sidepanel-open/content-script.js
+++ b/functional-samples/cookbook.sidepanel-open/content-script.js
@@ -2,7 +2,12 @@ const button = new DOMParser().parseFromString(
   '<button>Click to open side panel</button>',
   'text/html'
 ).body.firstElementChild;
+button.disabled = true;
 button.addEventListener('click', function () {
   chrome.runtime.sendMessage({ type: 'open_side_panel' });
 });
 document.body.append(button);
+
+chrome.runtime.sendMessage({ type: 'enable_tab_side_panel' }, () => {
+  button.disabled = false;
+});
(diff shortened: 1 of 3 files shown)


KirtiRamchandani:
Closing this earlier version in favor of #1695, which uses the preconfiguration approach for #1477 so the click handler only needs to call chrome.sidePanel.open().

PR #1689, opened 2026-05-25
Update global search sample region handling

### LINKED ISSUE #1653: "Global search" example seems outdated
The [`api-samples/contextMenus/global_context_search` example](https://github.com/GoogleChrome/chrome-extensions-samples/tree/main/api-samples/contextMenus/global_context_search) describes itself as, "letting a user switch between searching different countries' versions of Google", but that's no longer something that Google supports. While it's still possible to perform a regional Google search (see below), switching the domain for the request isn't how.

* I _**get**_ that it's just an example. 
* I _**get**_ that different search URLs **ARE** still generated based on the user's selections.
* I _**get**_ that it's still an effective demonstration of the `contextMenus` APIs that it's _meant_ to be demonstrating.

But, it's not really a great example anymore. The best examples demonstrate how to do something trivial but **useful**, letting the user experience _why_ these APIs exist and would be of interest to developers.

And because Google has globalized its service offerings, beyond the string displayed in the URL bar there is no longer any visible difference loading any of these URLs:

* https://www.google.ca/search?q=my+search
* https://www.google.co.jp/search?q=my+search
* https://www.google.ru/search?q=my+search

(They used to be different, long ago. I assume they still were, when the example was originally written. But no longer.)

In all cases today, Google will show the exact same search results, localized based on an approximate geolocation derived from the requesting IP address. Which makes the example seem pointless.

Perhaps a more current example would be useful. Like:

* Setting a Region parameter for the search. (`&cr=countryCA`  vs. `&cr=countryJP` vs. `&cr=countryRU` etc.) That _will_ have the effect that switching the TLD _used_ to have. (Although the results will still be in the user's presumed language, unless a corresponding `&lr=lang_en` / `&lr=lang_jp` / `&lr=lang_ru` / etc. parameter is also set.)
* Switching between different Google search categories via the `&udm=` parameter, though as that's largely undocumented (from official sources, at least) it may not be the best or more future-proof choice.
* Something else entirely?

KirtiRamchandani:
## Problem

The global context search sample still suggested that selecting a country switched to that country's Google domain. As noted in #1653, country-specific Google domains no longer provide a useful visible difference for this sample.

## Root cause

The sample encoded locale choice as Google Search hostnames such as `google.ca` or `google.co.jp`, so the UI described obsolete domain switching instead of an explicit search-region behavior.

## Solution

Use `https://www.google.com/search` for all searches and add the Google Search `cr=countryXX` country restriction parameter based on the selected context-menu item. Rename the sample UI/storage identifiers from TLDs to countries and refresh the README/manifest copy to describe country-restricted results.

## Tests run

- `npx eslint --no-warn-ignored api-samples/contextMenus/global_context_search/background.js api-samples/contextMenus/global_context_search/popup.js --quiet`
- `npx prettier --check api-samples/contextMenus/global_context_search/background.js api-samples/contextMenus/global_context_search/locales.js api-samples/contextMenus/global_context_search/popup.js api-samples/contextMenus/global_context_search/popup.html api-samples/contextMenus/global_context_search/manifest.json api-samples/contextMenus/global_context_search/README.md`
- `git diff --check HEAD~1..HEAD`
- Node smoke test with mocked `chrome.*` APIs verifying menu creation, `cr=countryCA` search URLs, and country storage updates

Fixes #1653

### DIFF
--- a/api-samples/contextMenus/global_context_search/README.md
+++ b/api-samples/contextMenus/global_context_search/README.md
@@ -1,10 +1,10 @@
 # chrome.contextMenus
 
-This sample demonstrates the `chrome.contextMenus` API by letting a user switch between searching different countries' versions of Google via a `contextMenu`.
+This sample demonstrates the `chrome.contextMenus` API by letting a user restrict Google Search results to pages from different countries via a `contextMenu`.
 
 ## Overview
 
-The extension uses `chrome.contextMenus.create()` to populate the context menu with locale options based on an options menu in the popup. A `chrome.contextMenus.onClicked.addListener()` event will open a specific locale's Google homepage when one of the extension's context menu options are clicked.
+The extension uses `chrome.contextMenus.create()` to populate the context menu with country options based on an options menu in the popup. A `chrome.contextMenus.onClicked.addListener()` event opens Google Search with a `cr=countryXX` query parameter when one of the extension's context menu options is clicked.
 
 ## Running this extension
 
--- a/api-samples/contextMenus/global_context_search/background.js
+++ b/api-samples/contextMenus/global_context_search/background.js
@@ -4,14 +4,14 @@
 
 // When you specify "type": "module" in the manifest background,
 // you can include the service worker as an ES Module,
-import { tldLocales } from './locales.js';
+import { countryLocales } from './locales.js';
 
 // Add a listener to create the initial context menu items,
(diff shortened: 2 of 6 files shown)


KirtiRamchandani:
Closing this in favor of the cleaned-up follow-up in #1697, which keeps the same issue focused but removes the extra unrelated file touch and has the latest verification. Thanks.

PR #1718, opened 2026-07-19
Text Replacer example

dotproto:
Adds a sample extension called "Text Replacer." As the name implies, this extension will replace a string in the page with a custom string. Currently supports 5 replacements.

This PR takes the place of #612, which I'm no longer able to directly update.


functional-samples/sample.text-replacer/manifest.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "Text Replacer",
+  "version": "1.0.0",
+  "manifest_version": 3,
+  "description": "Replace any a line of text on a page with another line of text.",
```
patrickkettner: nit: typo? "any a line"

functional-samples/sample.text-replacer/text-replacer/content.js
```diff
@@ -0,0 +1,47 @@
...
+  while (node = nodeIterator.nextNode()) {
+    for (let [find, replace] of replacements) {
+      node.nodeValue = node.nodeValue.replace(find, replace);
+    }
+  }
+}
+
+// Replace text on the page using a list of patterns loaded from storage
+chrome.storage.sync.get(['patterns'], function(data) {
+  textReplacer(data.patterns);
```
patrickkettner: if `patterns` hasnt been set yet, then you are passing `undefined` to textReplacer, which would throw because it passes it through to buildReplacementRegex, which will try to access `${undefined}.length` on line 15

functional-samples/sample.text-replacer/text-replacer/manifest.json
```diff
@@ -0,0 +1,35 @@
...
+  "version": "1.0.0",
+  "manifest_version": 3,
+  "description": "Replace any a line of text on a page with another line of text.",
+  "icons": {
+    "32": "icons/icon32.png",
+    "48": "icons/icon48.png",
+    "64": "icons/icon64.png"
+  },
+  "action": {
+    "default_title": "Show an alert",
```
patrickkettner: is this title on purpose?

functional-samples/sample.text-replacer/text-replacer/popup.js
```diff
@@ -0,0 +1,130 @@
...
+    });
+  });
+}
+
+/**
+ * Minimal promise wrapper for chrome.storage.sync.get().
+ *
+ * @param {string[]} keys  Array of keys to retrieve from storage.
+ */
+function load(keys) {
```
patrickkettner: `storage.sync.get` returns a promise natively, so this can be cleaned up

patrickkettner (commented):
Thanks a bunch, dude! couple of small things inline, and then 

- the 19px image isn't used (probably remove it?). Also, could you add a 128px image, as thats the size the store requests folks add.
- Could you add a readme link the sibling functional samples?
- your google account is still attached to the commit, can you run ` git commit --amend --no-edit && git push --force-with-lease` to clear it out so the CLA doesn't block?

functional-samples/sample.text-replacer/manifest.json
dotproto: Done.[](url)

functional-samples/sample.text-replacer/text-replacer/content.js
dotproto: This line doesn't exist in the current version (@rdcronin asked that we request the data from the background rather than directly accessing through storage here), but I've fixed the underlying bug.

functional-samples/sample.text-replacer/text-replacer/manifest.json
dotproto: Nope. I fixed this in the batch of issues identified #612.

functional-samples/sample.text-replacer/text-replacer/popup.js
dotproto: I've simplified this and other storage access code. All reads/writes are now routed through the background and we use native promises.

dotproto:
@patrickkettner, I think I've covered everything you request and all of the issues identified in #612.

functional-samples/sample.text-replacer/background.js
```diff
@@ -0,0 +1,54 @@
...
+  }
+});
+
+// Route all storage reads/writes through the background so we have a single
+// source of truth.
+chrome.runtime.onMessage.addListener((message, ) => {
+  switch (message.id) {
+    case GET_REPLACEMENTS_MESSAGE_ID:
+      // Fall back to an empty array 'patterns' is not set
+      return chrome.storage.sync.get({patterns: []});
```
patrickkettner: returning a promise is chrome 148+, so its still new-ish. Love using it! But probably set a min chrome version, or comment to that effect

functional-samples/sample.text-replacer/content.js
```diff
@@ -0,0 +1,53 @@
+// Copyright 2021 Google LLC
+//
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file or at
+// https://developers.google.com/open-source/licenses/bsd
```
patrickkettner: License needs to be updated to be the Apache header used in other files in the repo (with correct year)

functional-samples/sample.text-replacer/content.js
```diff
@@ -0,0 +1,53 @@
...
+// Wrap the in an immediately invoked function expression (IIFE) in order to
+// prevent local variables from polluting global scope
+(function () {
+  const GET_REPLACEMENTS_MESSAGE_ID = 'get-replacements';
+
+
+  function buildReplacementRegex(source) {
+    const output = [];
+    for (let i = 0; i < source.length; i++) {
+      if (!source[i]) { continue; }
```
patrickkettner: What happens if the saved replace value is undefined? Wouldnt that get injected between every word?

patrickkettner (commented):
couple nits! License comment applies to all files that have BSD at the moment

functional-samples/sample.text-replacer/content.js
dotproto: Done.

functional-samples/sample.text-replacer/content.js
dotproto: Do you mean if the value of the variable `replace` (line 16) is literally `undefined`? If so, yes, the `replace()` call made by `replaceText()` would coerced that value into the string `"undefined"` and that string would be inserted in the document. But in practice that should not happen.

The value saved to `patterns` key is assembled in `saveFormData()` from the `.value` properties on each of the `<input>` elements in the popup. Since these are text inputs, an empty value is still an empty string.

Still, we can be safe here and guard against it. I've added a nullish assignment check here so if the value is null or undefined, we use an empty string instead.

functional-samples/sample.text-replacer/background.js
dotproto: Added `minimum_chrome_version` to the manifest and a note in the readme about why it's there.

functional-samples/sample.text-replacer/content.js
patrickkettner: Sorry, I explained myself badly. I meant what if a value in the array is an empty string, like `['', 'XX']`. doing `(!source[i]?.[0])` would prevent it

foodstampou812 (approved):

functional-samples/sample.text-replacer/content.js
dotproto: Oh! You're totally right. Hm...`(!source[i]?.[0])` works, but I think it's too wonky and not obvious enough. This also made me look closer at the current implementation and I don't like `output[i] = [findExp, replace]` as a general pattern because it can create holey arrays.

I've revised this function to use a functional approach that is more robust and, hopefully, more clear to readers of all skill levels.

dotproto:
I suggest enabling [Code review limits](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/managing-repository-settings/managing-pull-request-reviews-in-your-repository#enabling-code-review-limits) on this repo to prevent drive-by approvals.

patrickkettner:
thanks!

PR #1784, opened 2026-08-27
Simplify the declarativeNetRequest URL redirect sample

### LINKED ISSUE #1783: Replace the declarativeNetRequest url-redirect sample with a simpler one
`api-samples/declarativeNetRequest/url-redirect` currently redirects real developer.chrome.com doc URLs, which is what made it confusing (and is why #1083 exists at all). Replace it with a simpler example, e.g. redirecting to example.com.

patrickkettner:
The five rules pointed at real developer.chrome.com pages, which made the sample look like broken navigation rather than a redirect demo. Replaced them with one rule redirecting example.com/original to example.com/redirected.

Fixes #1783


oliverdunk (approved):
Much better - thanks!

